PR #4, opened 2022-01-05
run: hook server which can parse `DNSbackend` and `listen ips`

flouthoc:
Implement `aardvark-dns --file <config> run` to actually parse and run server

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -1,22 +1,27 @@
 //! Runs the aardvark dns server with provided config
+use crate::server::serve;
 use clap::{self, Clap};
 use log::debug;
-use core::fmt::Error;
+use std::io::Error;
 
 #[derive(Clap, Debug)]
-pub struct Run {
-}
+pub struct Run {}
 
 impl Run {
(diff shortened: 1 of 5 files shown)


flouthoc:
@mheon Merging this one as well since its a synthetic change.

src/commands/run.rs
```diff
@@ -1,22 +1,27 @@
...
 
 impl Run {
     /// The run command runs the aardvark-dns server with the given configuration.
     pub fn new() -> Self {
-        Self {
-        }
+        Self {}
     }
 
     pub fn exec(&self, input_file: String) -> Result<(), Error> {
```
mheon, line 16: It's not a file, it's a directory. We accept a context directory containing 1 or more config files.

flouthoc: oh I'll rename the variable in the next PR. However Its treated as a directory so it should not break.

flouthoc: @mheon Fixed in a commit specified here: https://github.com/containers/aardvark-dns/pull/6

PR #6, opened 2022-01-06
parse: `ipv4` or `ipv6` field could be empty as well so use `Option<`

flouthoc:
* First line of config could contain single ip or multiple separated by, so commit fixes this edge case.
* Ctr entries could have a blank ip field, commit deals with that.
* End properly when file ends or ignore blank lines

___
Second commit fixes issue addressed here: https://github.com/containers/aardvark-dns/pull/4#discussion_r778852498


flouthoc:
@mheon I am merging this one as i need to push dns core.

PR #7, opened 2022-01-06
dns: add implementation for aardvark's `core dns` server

flouthoc:
Adds implementation for Aardvark's core dns server in
`src/dns/coredns.rs`

Following core will will used by Aardvark's server to handle actual dns
requests and responses. 

This will be used in `src/server/serve.rs` with actual business logic.


flouthoc:
@baude @mheon Review should start from file `src/dns/coredns.rs`

But i might merge this PR since i have to complete interface between `server` <-> `core dns`

PR #8, opened 2022-01-06
server: add `A/AAAA` records from config to coredns `in-memory` authority

flouthoc:
* Integrate server with coredns so it can add `A/AAAA` records
* Config modify to return `HashMap<networkname, listenips>` from `Vec<listenip>`
    
    There are some hardcodings so this only works with as of now with
    https://github.com/containers/aardvark-dns/blob/main/src/test/config/podman
    
* Add a try out config


mheon:
Why modify the config? We shouldn't need to know what network a result is from, just that there were a number of results for the given query. We return every result that is returned by the query in that Vec<listenip> not just the ones in the specific network that is asking

flouthoc:
@mheon No config is not changed, just returning `network name` with `listen ip`. So I that i can prevent creating copies of `container <--> ip` in authority store. The config is still intact.

PR #9, opened 2022-01-07
dns: allow `underscore` in record origin, fallback to `utf8` instead of `IDNA` parsing

flouthoc:
We need dns server to be able to understand `underscore` in record
names. So fallback to `utf8` if `IDNA` fails and try using record name
anyhow.

References:
https://datatracker.ietf.org/doc/html/rfc952
https://www.iana.org/assignments/idna-tables-6.3.0/idna-tables-6.3.0.xhtml
https://datatracker.ietf.org/doc/html/rfc5890#section-2.3.1

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -69,7 +69,10 @@ impl CoreDns {
     }
 
     pub fn update_record(&mut self, name: &str, addr: IpAddr, ttl: u32) {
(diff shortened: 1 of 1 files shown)


flouthoc:
@baude @mheon FYI

flouthoc:
Merging this one. As its already discussed and we need this.

PR #10, opened 2022-01-09
dns: hook `DNSBackend` for `A/AAAA` records lookup instead of using `Authority` records

flouthoc:
Configure coredns to use lookup logic directly from `DNSBackend` instead
of using `Authority` records.

This prevents creating copies of `A/AAAA` record entries and we follow
desired logic for lookup.


flouthoc:
@mheon Following prevents creating copies and makes sure that we use lookup logic from `DNSBackend` as it is.

PR #12, opened 2022-01-11
server, coredns: Refresh DNS servers on `SIGHUP` with global `Mutex`

flouthoc:
* Implements global `kill_switch` for Coredns with a shared `Mutex`
* Server: Perform reload of Backend on `SIGHUP`
* Accept aardvark port from cli. Adds `--port` to global flag.


flouthoc:
Hi @edsantiago Following PR adds a tryout example to run `aardvark` on local using config https://github.com/containers/aardvark-dns#run-example 

I am adding more docs in a followup.

PR #14, opened 2022-01-12
server: spin servers for `IPv6` dynamically from config

flouthoc:
Aardvark config can contain list of both `IPv4` and `IPv6`.
Following PR ensures that we spin servers for `IPv6` as well
without ignoring.


flouthoc:
@mheon @baude This should close `IPv6` spin ups.

PR #17, opened 2022-01-17
Add rpkg template for COPR autobuild

lsm5:
https://lists.fedoraproject.org/archives/list/copr-devel@lists.fedorahosted.org/message/3JORYRKDWFMJSR35Z4LIKDEXH2T5263H/

Also comment out docs install step in Makefile

@flouthoc @baude @mheon @Luap99 @rhatdan PTAL.

Btw, I'm not the biggest fan of install targets depending on build targets. Would it be ok to have them separate? This would mean `make && make install` for people building from source, but it would also mean no extra `-nobuild` targets like the mess we have in podman.


lsm5:
Let me know what the aardvark <-> netavark <-> podman install deps should look like. I can add them to the PRs.

baude:
podman4 should require netavark, netavark should require aardvark

baude:
LGTM

lsm5:
> podman4 should require netavark, netavark should require aardvark

i'll update PRs for podman and netavark. This one should be good to go.

flouthoc (approved):

PR #18, opened 2022-01-18
server, daemon: Kill server when not needed and make it `daemonless`.

flouthoc:
Lets not run aardvark as a daemon.

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -72,6 +72,13 @@ fn core_serve_loop(_config_path: &str, port: u32) -> Result<(), std::io::Error> 
             let mut thread_handles = vec![];
             let kill_switch = Arc::new(Mutex::new(false));
 
+            // kill server if listen_ip's are empty
+            if listen_ip_v4.is_empty() && listen_ip_v6.is_empty() {
+                //no configuration found kill the server
+                println!("No configuration found stopping the sever");
+                process::exit(0);
+            }
+
             // Prevent memory duplication: since backend is immutable across threads so create Arc and share
             let shareable_arc = DNSBackendWithArc {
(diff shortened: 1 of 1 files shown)


flouthoc:
FYI @baude @mheon @Luap99 @rhatdan As discussed we don't need aardvark to be daemon so server automatically kills itself when not needed and `netavark` will restart it whenever its needed. 

Startup logic here: https://github.com/containers/netavark/pull/162

src/server/serve.rs
```diff
@@ -72,6 +72,13 @@ fn core_serve_loop(_config_path: &str, port: u32) -> Result<(), std::io::Error>
             let mut thread_handles = vec![];
             let kill_switch = Arc::new(Mutex::new(false));
 
+            // kill server if listen_ip's are empty
+            if listen_ip_v4.is_empty() && listen_ip_v6.is_empty() {
+                //no configuration found kill the server
+                debug!("No configuration found stopping the sever");
```
Luap99, line 78: Can we do this at a level that is visible by default.

flouthoc: @Luap99 I am sorry but I did not understand what do we mean by `visible` by default. Could you help me understand it.

We want to stop aardvark when no listen ip's are present. That is the intention of this PR.

flouthoc: Did you mean make it configurable ?

Luap99: debug lines are not printed unless debug level is set, I want to see this line by default (when no log level is explicitly set). Just to make sure it exited intentionally and did not crashed for some other reason.

flouthoc:
I am merging this since nit is resolved and this is fairly simple change.

PR #26, opened 2022-01-20
Makefile: install target shouldn't call build

lsm5:
This would imply `make && make install` for people building from source.
It would also save the Makefile from getting polluted with `-nobuild`
targets.

@flouthoc @baude @mheon @Luap99 @rhatdan PTAL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ clean:
 #	$(MAKE) -C docs
 
 .PHONY: install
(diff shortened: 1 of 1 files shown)


flouthoc (approved):

lsm5:
@baude @mheon PTAL

mheon:
LGTM

PR #27, opened 2022-01-21
Fix leaking pid file fd and remove pid file on exit 

Luap99:
Rust closes the fd only when it leaves the scope, since this is
the main loop it will never happen so we have to manually close it.
Also we should not leak the pid file on exit.


Luap99:
@baude @flouthoc PTAL

flouthoc (approved):

PR #28, opened 2022-01-21
server, hotfix: stop holding child threads as they are not relevent

flouthoc:
Hotfix: Just rely on kill switch and don't wait for child threads.


flouthoc:
@baude Merge me please if you have verified it with tests

PR #29, opened 2022-01-21
coredns: set `QR` to `1` bit for response messages

flouthoc:
It seems we don't set QR bit to `1` for dns response lets do that.


baude:
LGTM

PR #32, opened 2022-01-25
dns, server: don't trigger kill_switch inside receiver block instead move it to parent `event loop`.

flouthoc:
Why ?
    It seems there is a race if interface is removed from below before we stop event loop and closing
    sockets won't stop the event loop. So lets de-couple thread killing from any network code.
    
    Solution:
    Receiver is already async so leverage that and move kill_switch to an
    event loop which is not dependent on receiver.
    
    Following solution makes sure that our thread management is not
    dependent on any networking code. If we want to kill the thread we can
    kill it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ else
 endif
 
 .PHONY: all
-all: vendor build
+all: build
 
 bin:
 	mkdir -p $@
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -4,6 +4,7 @@ use futures_util::StreamExt;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::{Arc, Mutex};
 use tokio::net::UdpSocket;
+use tokio::time::{self, Duration};
 use trust_dns_client::{client::AsyncClient, proto::xfer::SerialMessage, rr::Name};
 use trust_dns_proto::{
     op::{Message, MessageType},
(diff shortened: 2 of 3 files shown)


flouthoc:
How to test this ? Use following commit https://github.com/baude/aardvark-dns/commit/98148bbd32f42578983b66525aea2fced716f88e

with diff since aardvark does not persists pid after https://github.com/containers/aardvark-dns/pull/27
```diff
diff --git a/test/100-basic-name-resolution.bats b/test/100-basic-name-resolution.bats
index 0456780..b5d5aee 100644
--- a/test/100-basic-name-resolution.bats
+++ b/test/100-basic-name-resolution.bats
@@ -33,7 +33,7 @@ load helpers
 	create_container "$config_a1"
 	a1_pid=$CONTAINER_NS_PID
 	cat /tmp/aardvark-dns/podman1
-	cat /tmp/aardvark-dns/aardvark.pid
+	#cat /tmp/aardvark-dns/aardvark.pid
 
 	# container a2
 	create_config $(random_string 64) "atwo" "$subnet_a" "a2" "2a"
@@ -43,7 +43,7 @@ load helpers
 	a2_pid=$CONTAINER_NS_PID
 
     cat /tmp/aardvark-dns/podman1
-	cat /tmp/aardvark-dns/aardvark.pid
+	#cat /tmp/aardvark-dns/aardvark.pid
 	# can a1 resolve a2's hostname
 	echo "------------------"
 	#echo "$gw // $ip_a1 // $ip_a2"
@@ -54,4 +54,4 @@ load helpers
 	#run_in_host_netns ps aux | grep aardvark-dns
 	#run_in_container_netns "$a2_pid" "dig" "+short" "aone" "@$gw"
 
-}
\ No newline at end of file
+}
```

flouthoc:
I am merging this since I think upstream should be in working state. If anybody plays around with test.

flouthoc:
@baude @Luap99 @mheon @Luap99 Upstream and https://github.com/containers/aardvark-dns/pull/32#issuecomment-1020883941 should fix the issue.

PR #35, opened 2022-01-26
add version command

Luap99:
Add version command to get the version, commit sha, build time and the
rust target triple.

Also update clap to 3.0.12
version 3 has some breaking changes so we have to fix that


mheon:
LGTM

PR #38, opened 2022-01-27
coredns: skip proxy if `AARDVARK_NO_PROXY` is found and return `NXDOMAIN`

flouthoc:
By default aardvark forwards request to configured resolver, but in
testing scenarios no host env is available so allow users to bypass
proxy if they need to.

if env contains `AARDVARK_NO_PROXY=true` aardvark returns `NXDOMAIN` if no entry is found in
DNSBackend instead of forwarding it to configured resolver.

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -1,20 +1,20 @@
 use crate::backend::DNSBackend;
 use crate::backend::DNSResult;
 use futures_util::StreamExt;
+use log::{debug, error, trace, warn};
+use std::env;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::{Arc, Mutex};
 use tokio::net::UdpSocket;
 use tokio::time::{self, Duration};
 use trust_dns_client::{client::AsyncClient, proto::xfer::SerialMessage, rr::Name};
 use trust_dns_proto::{
-    op::{Message, MessageType},
+    op::{Message, MessageType, ResponseCode},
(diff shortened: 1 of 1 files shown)


src/dns/coredns.rs
```diff
@@ -227,6 +227,14 @@ impl CoreDns {
                                 reply(sender, src_address, &req);
                             } else {
                                 debug!("Not found, forwarding dns request for {:?}", name);
+                                match env::var("AARDVARK_NO_PROXY") {
```
Luap99, line 230: Can we lookup the env var once and store it as bool instead of looking it up for every single request.

src/dns/coredns.rs
```diff
@@ -227,6 +227,14 @@ impl CoreDns {
...
+                                    Ok(_) => {
+                                        let mut nx_message = req.clone();
+                                        nx_message.set_response_code(ResponseCode::NXDomain);
+                                        reply(sender.clone(), src_address, &nx_message).unwrap();
```
Luap99, line 234: Should there be a return here or something, otherwise we still call `tokio::spawn(async move { ...`  below

flouthoc: fixed

src/dns/coredns.rs
flouthoc: fixed.

Luap99 (commented):
LGTM

Issue #46, opened 2022-01-31
aardvark-dns consuming entire cpu


baude:
Reproducer:

$ podman system reset --force
$ podman create network
$ podman run -dt --rm --name baude --network podman1 alpine top
$ podman run -it --rm --network podman1 ping baude
cntrl-c after it pings
.... wait

strace shows:

futex(0x7f99f87ac910, FUTEX_WAIT_BITSET|FUTEX_CLOCK_REALTIME, 130307, NULL, FUTEX_BITSET_MATCH_ANY

flouthoc:
@baude Thanks above PR should close the issue and makes sure that `aardvark` uses almost negligible CPU and memory.

PR #48, opened 2022-02-01
server: use `async_broadcast` channel instead of delay in event loop

flouthoc:
We are using all most all of CPU with the approach where we are trying
to read value of kill switch in the event loop via tokio::time::Sleep. This approach looks bad.

Instead start listening to async broadcast channel as a kill switch
mechanism, this will make sure that we don't eat CPU for no reason.

Following also adds and uses a new crate: `async_broadcast` which is needed for mpmc ( multi-producer, multi-consumer ) broadcasting across threads.

Closes: https://github.com/containers/aardvark-dns/issues/46


flouthoc:
@baude Please verify the patch and merge-me.

src/dns/coredns.rs
```diff
@@ -18,14 +17,15 @@ use trust_dns_proto::{
...
-    backend: Arc<DNSBackend>,      // server's data store
-    kill_switch: Arc<Mutex<bool>>, // global kill_swtich
-    filter_search_domain: String,  // filter_search_domain
+    name: Name,                          // name or origin
+    address: IpAddr,                     // server address
+    port: u32,                           // server port
+    authority: InMemoryAuthority,        // server authority
+    cl: AsyncClient,                     //server client
+    backend: Arc<DNSBackend>,            // server's data store
+    kill_switch: Arc<Mutex<bool>>,       // global kill_swtich
```
giuseppe, line 26: typo in the comment

flouthoc: thanks fixed the typo.

baude:
LGTM, appears to fix the problem

Issue #47, opened 2022-01-31
Aardvark-dns should use the systems /etc/resolv.conf when forwarding DNS requests.


rhatdan:
It needs to use the dns-options, dns nameservers and the dns-searches.

rhatdan:
On Red Hat internal network, DNS resolution is only allowed from the internal dns servers, not from public ones.

flouthoc:
@rhatdan Thanks above PR should close the issue.

PR #49, opened 2022-02-01
coredns: use host's `/etc/resolv.conf` if available as forwarder

flouthoc:
Use host's `/etc/resolv.conf` if available and forward request to
host's resolver.

If everything fails try using configured global resolver.

Following PR adds new crate: `trust-dns-resolver` automatically use host's resolver 

Closes: https://github.com/containers/aardvark-dns/issues/47


flouthoc:
@baude @Luap99 @rhatdan This should fix issues if for cases where private VPN blocks other resolver.

baude:
LGTM, but I would want @rhatdan to verify ...

src/dns/coredns.rs
```diff
@@ -238,8 +239,51 @@ impl CoreDns {
...
+                                            }
+                                        }
+
+                                        // using default from hosts's /etc/resolv.conf failed
+                                        // as a last resort try using global configured forwarder
+                                        // default global forwarder is 1.1.1.1
+                                        if !did_reply {
+                                            if let Some(resp) = forward_dns_req(client, req.clone()).await {
+                                                reply(sender.clone(), src_address, &resp).unwrap();
+                                            }
```
Luap99, line 286: We should not fallback to a hardcoded resolver IMO. If /etc/resolv.conf does not work than this is a user error.

src/dns/coredns.rs
```diff
@@ -238,8 +239,51 @@ impl CoreDns {
...
+                                        if let Ok(resolver) = TokioAsyncResolver::tokio_from_system_conf() {
+                                            if let Ok(lookup_ip) = resolver.lookup_ip(name.as_str()).await {
+                                            let record_name: Name = Name::from_str_relaxed(name.as_str()).unwrap();
+                                                for response_ip in lookup_ip {
+                                                        match response_ip {
+                                                            IpAddr::V4(ipv4) => {
+                                                                req.add_answer(
+                                                                     Record::new()
+                                                                        .set_name(record_name.clone())
+                                                                        .set_ttl(86400)
```
Luap99, line 254: We should keep the original ttl no?

src/dns/coredns.rs
baude: agree

PR #53, opened 2022-02-02
coredns: forward request as it-is when using host's resolver i.e `/etc/resolv.conf`

flouthoc:
Using trust-dns-resolver forces us to read ip , while we don't need to
do it since we should be acting as a proxy. Following PR makes sure that
we forward DNS response as it-is from host's resolver.

* Fixes the issue which was discussed where we were not honoring actual
TTL.


flouthoc:
@mheon @Luap99 @baude @rhatdan This fixes the `TTL` issue which was reported yesterday and also removes `trust-dns-resolver` which is not solving our use-case.

src/dns/coredns.rs
```diff
@@ -222,39 +242,24 @@ impl CoreDns {
...
-                                                                        .set_name(record_name.clone())
-                                                                        .set_ttl(86400)
-                                                                        .set_rr_type(RecordType::AAAA)
-                                                                        .set_dns_class(DNSClass::IN)
-                                                                        .set_rdata(RData::AAAA(ipv6))
-                                                                        .clone(),
-                                                                );
-                                                            }
-                                                        }
+                                        for nameserver in nameservers {
```
mheon, line 248: If all servers fail, do we respond at all?

flouthoc: @mheon It should time out. It was decided by everyone in the last PR.

mheon: Alright, seems reasonable.

mheon:
Changes LGTM

flouthoc:
@Luap99 @baude Please verify patch and merge me.

Luap99 (approved):
LGTM

PR #54, opened 2022-02-02
Support reverse lookups via in-memory lookup table

mheon:
Updated version of #39 which avoids expensive traversals of maps.


src/dns/coredns.rs
```diff
@@ -138,6 +138,75 @@ impl CoreDns {
...
+                                    ptr_lookup_ip = &ptr_lookup_ip[..len-limit];
+                                    let ip_octs: Vec<&str> = ptr_lookup_ip.split('.').collect();
+                                    let reverse_ip: Vec<&str> = ip_octs.into_iter().rev().collect();
+                                    reverse_string = reverse_ip.join(".");
+                                    reverse_string = (&reverse_string[1..reverse_string.len()]).to_string();
+                                    ptr_lookup_ip = &reverse_string;
+                                }
+                                // reverse the ip
+                                trace!("perform lookup reverse lookup for ip: {:?}", ptr_lookup_ip.to_owned());
+                if let Ok(lookup_ip) = ptr_lookup_ip.parse() {
```
mheon, line 192: I know the indentation here is whacky but `make validate` on my system blows up if I don't use it

flouthoc: @mheon Its because `cargo fmt` is not capable of detecting indentation inside `macros` which we are using for `select`. You can manually indent like https://github.com/containers/aardvark-dns/pull/39 and I'll add rules `clippy` and `fmt` later.

Otherwise indentation really looks off.

flouthoc:
@mheon I tried running this PR but reverse lookup does not works with this. I am suspecting something up with the lookup logic. Could you also try running the PR.

Should we merge original first and only do only optimization in this PR: https://github.com/containers/aardvark-dns/pull/39 doing so would make sure we can unblock reverse lookup tests in CI and since we have finite entries I don't think performance is a issue here.

But I am fine with anything.

src/backend/mod.rs
```diff
@@ -85,4 +89,24 @@ impl DNSBackend {
 
         DNSResult::Success(results)
     }
+
+    // reverse lookup must return a single name resolved via mapping
+    pub fn reverse_lookup(&self, requester: &IpAddr, lookup_ip: &IpAddr) -> String {
```
flouthoc, line 94: @mheon So one more thing i figured out is we must return `Vec<String>` cause currently list also contains `container id` and `name` so we might end up returning `containerid`, I have also made change in previous PR if you need any reference: https://github.com/containers/aardvark-dns/pull/39

flouthoc:
@mheon If you are caught with podman stuff do you want me to take over this PR ?

src/backend/mod.rs
```diff
@@ -85,4 +89,24 @@ impl DNSBackend {
...
+    // reverse lookup must return a single name resolved via mapping
+    pub fn reverse_lookup(&self, requester: &IpAddr, lookup_ip: &IpAddr) -> Vec<String> {
+        let nets = match self.ip_mappings.get(requester) {
+            Some(n) => n,
+            None => return vec![],
+        };
+
+        for net in nets {
+            match self.reverse_mappings.get(net) {
+                Some(n) => match n.get(lookup_ip) {
```
cgwalters, line 102: I'd use `if let Some(n) = self.reverse_mappings.get(net) {` if your `None` case is just `{}`.

src/backend/mod.rs
```diff
@@ -85,4 +89,24 @@ impl DNSBackend {
...
+                    Some(i) => return i.clone(),
```
cgwalters, line 103: Is there a reason not to borrow here, i.e. return `Vec<&str>` and have the caller do the clone if they need to?

src/dns/coredns.rs
```diff
@@ -138,6 +138,77 @@ impl CoreDns {
                                 self.kill_switch.lock().unwrap()
                             );
 
+                            // if record type is PTR try resolving early and return if record found
+                            if record_type == RecordType::PTR {
```
cgwalters, line 142: This is very deeply nested code, seems like some of this could be broken out into helper fuctions that could also be unit tested e.g.

src/dns/coredns.rs
```diff
@@ -138,6 +138,77 @@ impl CoreDns {
...
+                                        if c == '.' {
+                                            dots += 1;
+                                        }
+                                    }
+                                    if ptr_lookup_ip.matches('.').count() == 4 {
+                                        limit = 1;
+                                    }
+                                    ptr_lookup_ip = &ptr_lookup_ip[..len-limit];
+                                    let ip_octs: Vec<&str> = ptr_lookup_ip.split('.').collect();
+                                    let reverse_ip: Vec<&str> = ip_octs.into_iter().rev().collect();
```
cgwalters, line 185: I think this is just:
```
                                   let reverse_ip = ptr_lookup_ip.split('.).rev().collect();
```
no?

flouthoc:
I tested the PR manually with podman, works for me.

mheon:
Seems to work now, but the tests need to be rewritten as we return not just the container name, but aliases as well. Also need to address comments from Colin.

src/dns/coredns.rs
```diff
@@ -158,6 +158,57 @@ impl CoreDns {
...
+                                    ptr_lookup_ip = split.join(":");
+                                } else {
+                                    ptr_lookup_ip = String::from("");
+                                }
+
+                                trace!("perform lookup reverse lookup for ip: {:?}", ptr_lookup_ip.to_owned());
+                                // We should probably log malformed queries, but for now if-let should be fine.
+                                if let Ok(lookup_ip) = ptr_lookup_ip.parse() {
+                                    let reverse_lookup = self.backend.reverse_lookup(&src_address.ip(), &lookup_ip);
+                                    if !reverse_lookup.is_empty() {
```
cgwalters, line 193: This seems like an argument for `fn reverse_lookup() -> Option<Vec<String>>`.


Then inside the function instead of returning an empty vec it's just `None`, and this is an `if let Some()` etc.

mheon: Fair. I'd also still like to make it return &str. But these are both things for Monday.

mheon: (For a final return type of `Option<Vec<&str>>`)

cgwalters: >  But these are both things for Monday.

Definitely, just glanced at the code out of curiosity; nothing I'm saying here should be considered to block anything!  And have a good weekend!

baude:
```
diff --git a/test/500-reverse-lookups.bats b/test/500-reverse-lookups.bats
index e30d708..9453dab 100644
--- a/test/500-reverse-lookups.bats
+++ b/test/500-reverse-lookups.bats
@@ -27,7 +27,11 @@ load helpers
 
        # Resolve IPs to container names
        dig_reverse "$a1_pid" "$a2_ip" "$gw"
-       assert "atwo"
+       assert "$output" =~ "atwo"
+       assert "$output" =~ "a2"
+       assert "$output" =~ "2a"
        dig_reverse "$a2_pid" "$a1_ip" "$gw"
-       assert "atwo"
+       assert "$output" =~ "aone"
+       assert "$output" =~ "a1"
+       assert "$output" =~ "1a"
 }

````

mheon:
Thanks, will get those applied.

PR #57, opened 2022-02-03
dns: return `NXDOMAIN` if request was for internal entry and entry is not there.

flouthoc:
If request is for internal entry and we dont have internal entry
return `NXDOMAIN`.

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -217,7 +217,9 @@ impl CoreDns {
                                 reply(sender, src_address, &req);
                             } else {
                                 debug!("Not found, forwarding dns request for {:?}", name);
-                                if no_proxy {
+                                let request_name = name.as_str().to_owned();
+                                let filter_search_domain_ndots = self.filter_search_domain.clone() + ".";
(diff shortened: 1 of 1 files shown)


src/dns/coredns.rs
```diff
@@ -217,7 +217,8 @@ impl CoreDns {
                                 reply(sender, src_address, &req);
                             } else {
                                 debug!("Not found, forwarding dns request for {:?}", name);
-                                if no_proxy {
+                                let request_name = name.as_str().to_owned();
+                                if no_proxy || request_name.ends_with("dns.podman") || request_name.ends_with("dns.podman.") || request_name.matches('.').count() == 1  {
```
Luap99, line 221: I thought the dns search domain is configurable so we should use the same value here.

flouthoc: @Luap99 Yes actually I was just testing if CI fails so forgot editing it. Fixed it now.

flouthoc:
@baude It seems we don't have a check for bad entries in CI that's why all the tests are passing. Should i add few tests cases ?

mheon:
LGTM, but a new test would be a good idea

baude:
> LGTM, but a new test would be a good idea

we dont need a new test, i thijnk it is there, and im looking, hang tight boys

PR #58, opened 2022-02-03
Add "IBM" use case 1/2

baude:
Verify that one of the IBM use cases where we have three networks (A, B,
C).  And the a container is created on A and one on C, referred to as
A1 and B1.  Then A1 and C1 do a network connect and join network B and
they each add and alias.


mheon:
If tests pass, LGTM

test/helpers.bash
```diff
@@ -512,4 +512,11 @@ function dig_reverse() {
...
     #run_in_container_netns "$1" "dig" "-x" "$2" "+short" "@$3"
     run_in_container_netns "$1" "nslookup" "$2" "$3"
+}
+
+#
+# Pops a number of arguments from CONTAINER_CONFIGS
+#
+function remove_container_configs() {
+    unset CONTAINER_CONFIGS[${#CONTAINER_CONFIGS[@]}-$1]
 }
```
Luap99, line 522: Missing newline

test/300-three-networks.bats
```diff
@@ -4,3 +4,185 @@
...
+	assert "$b1_ip"
+
+	# c1 should be able to resolve aone
+	dig "$c1_pid" "aone" "$c_gw"
+	assert "$b1_ip"
+
+	# b2c1 should be able to resolve aone_nw alias
+	dig "$c1_pid" "aone_nw" "$c_gw"
+	assert "$b1_ip"
+}
```
Luap99, line 188: missing newline

test/helpers.bash
baude: I removed this function in favor of just resetting the array as you suggested.

flouthoc:
@baude It seems you did some shuffling in tests and now CI is breaking.

baude:
all set

flouthoc:
LGTM

PR #62, opened 2022-02-04
check for nxdomain

baude:
when dig'ing at a host that should not resolve, we now also check with
the host command to make sure we get a NXDOMAIN response.


flouthoc:
LGTM

PR #65, opened 2022-02-04
tests: add slirp4netns to connect the host netns

Luap99:
To test upstream resolvers we need a connection to the internet. We can
use slirp4netns for this. Also create a new mount namespace so that we
can bind mount a correct resolv.conf which points to the slirp4netns
resolver.


Luap99:
I had to wait for new CI images with slirp4netns. I think this should work now let me rebase and see

Luap99:
@baude @mheon @flouthoc PTAL

flouthoc (approved):
LGTM

PR #67, opened 2022-02-07
dns: A few minor patches

cgwalters:
dns: Return fully borrowed value

This is even simpler and more efficient.

---

dns: Use triple slash function documentation

A bit better best practice.

---

dns: Use `filter_map` to lower nesting

Combinators versus explicit matching can be a bit of a taste thing;
in this case I find `filter_map()` avoids nesting and makes things
a bit clearer.

---

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -91,7 +91,7 @@ impl DNSBackend {
     }
 
     // reverse lookup must return a single name resolved via mapping
-    pub fn reverse_lookup(&self, requester: &IpAddr, lookup_ip: &IpAddr) -> Option<Vec<&str>> {
+    pub fn reverse_lookup(&self, requester: &IpAddr, lookup_ip: &IpAddr) -> Option<&Vec<String>> {
         let nets = match self.ip_mappings.get(requester) {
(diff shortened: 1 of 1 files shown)


mheon:
LGTM. God, I missed having proper functional programming when I was using Go - `map()` is incredible

cgwalters:
Here's an even more combinator way:

```
index a3f0b2d..3bd24b0 100644
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -97,12 +97,8 @@ impl DNSBackend {
             None => return None,
         };
 
-        for ips in nets.iter().filter_map(|v| self.reverse_mappings.get(v)) {
-            if let Some(names) = ips.get(lookup_ip) {
-                return Some(names);
-            }
-        }
-
-        None
+        nets.iter()
+            .filter_map(|v| self.reverse_mappings.get(v))
+            .find_map(|ips| ips.get(lookup_ip))
     }
 }
```

(Though, I had to convince myself it has the same semantics, and there aren't unit tests, and this is about my limit for drive by PR time :wink: )

flouthoc (approved):

PR #72, opened 2022-02-09
server: remove any direct panics from server

flouthoc:
Never stop core server is any thread under the hood panics.


flouthoc:
I have remove some `unwraps` from dns as well. So hold this PR till i add another commit.

PR #73, opened 2022-02-10
dns: Trim server name if they are too long and remove `unwraps`

flouthoc:
* Trust dns have a limit on `server_names` so automatically trim server name.
* Remove pending `unwraps` from server.


flouthoc:
@baude PTAL

src/dns/coredns.rs
```diff
@@ -256,7 +265,14 @@ impl CoreDns {
                                     }
                                 }
                             }
-                            let record_name: Name = Name::from_str_relaxed(name.as_str()).unwrap();
+                            let record_name: Name = match Name::from_str_relaxed(name.as_str()) {
+                                Ok(name) => name,
+                                Err(e) => {
+                                    // log and continue server
+                                    debug!("Error while parsing record name: {:?}", e);
```
baude, line 272: should this be an error vs debug?

baude:
one nit, else LGTM

src/dns/coredns.rs
flouthoc: We don't have any error message but we received `None`. But I agree this should be on `error!`

flouthoc:
@baude PTAL again.

Luap99:
Please add a test with a long network name to prevent regressions.

test/100-basic-name-resolution.bats
```diff
@@ -17,6 +17,18 @@ load helpers
 	assert "$ip_a1"
 }
 
+@test "basic container - dns itself with long network name" {
+	subnet_a=$(random_subnet 5)
+	create_config "bignetttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttworkname" $(random_string 64) "aone" "$subnet_a" "a1" "1a"
```
Luap99, line 22: Please make the name a variable and use this instead of having the full name three times, this is unreadable

test/100-basic-name-resolution.bats
```diff
@@ -17,6 +17,19 @@ load helpers
 	assert "$ip_a1"
 }
 
+@test "basic container - dns itself with long network name" {
+	subnet_a=$(random_subnet 5)
+	long_name="bignetttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttworkname"
+	create_config $(long_name) $(random_string 64) "aone" "$subnet_a" "a1" "1a"
```
Luap99, line 23:
```suggestion
	create_config "$long_name" $(random_string 64) "aone" "$subnet_a" "a1" "1a"
```

test/100-basic-name-resolution.bats
```diff
@@ -17,6 +17,19 @@ load helpers
...
+	config_a1=$config
+	ip_a1=$(echo "$config_a1" | jq -r .networks."$long_name".static_ips[0])
```
Luap99, line 25:
```suggestion
	ip_a1=$(echo "$config_a1" | jq -r .networks.$long_name.static_ips[0])
```

test/100-basic-name-resolution.bats
```diff
@@ -17,6 +17,19 @@ load helpers
...
+	gw=$(echo "$config_a1" | jq -r .network_info."$long_name".subnets[0].gateway)
```
Luap99, line 26:
```suggestion
	gw=$(echo "$config_a1" | jq -r .network_info.$long_name.subnets[0].gateway)
```

test/100-basic-name-resolution.bats
```diff
@@ -17,6 +17,19 @@ load helpers
 	assert "$ip_a1"
 }
 
+@test "basic container - dns itself with long network name" {
+	subnet_a=$(random_subnet 5)
+	long_name="podman1111111"
```
Luap99, line 22: Well this is no longer a name, as I said before the network name should be long but interface name must be less than 16 chars.

flouthoc: Made it longer. But netavark blows up early on my local.

Issue #59, opened 2022-02-03
DO NOT PANIC!


Luap99:
There are way to many uses of `.unwrap()` here:
```
$ grep -R "unwrap()" src/
src/backend/mod.rs:            if name.len() > 0 && name.chars().last().unwrap() == '.' {
src/config/mod.rs:                if cfg.path().file_name().unwrap() == constants::AARDVARK_PID_FILE {
src/config/mod.rs:                        new_ctr_ips.push(IpAddr::V4(entry.v4.unwrap()));
src/config/mod.rs:                        new_ctr_ips.push(IpAddr::V6(entry.v6.unwrap()));
src/dns/coredns.rs:        let name: Name = Name::parse(name, None).unwrap();
src/dns/coredns.rs:        let origin: Name = Name::from_str_relaxed(name).unwrap();
src/dns/coredns.rs:                    match v.unwrap() {
src/dns/coredns.rs:                            let (name, record_type, mut req) = parse_dns_msg(msg).unwrap();
src/dns/coredns.rs:                                self.kill_switch.lock().unwrap()
src/dns/coredns.rs:                                                    request_name = request_name.strip_suffix(&self.filter_search_domain).unwrap().to_string();
src/dns/coredns.rs:                                                    request_name = request_name.strip_suffix(&filter_domain_ndots_complete).unwrap().to_string();
src/dns/coredns.rs:                            let record_name: Name = Name::from_str_relaxed(name.as_str()).unwrap();
src/dns/coredns.rs:                                    reply(sender.clone(), src_address, &nx_message).unwrap();
src/dns/coredns.rs:                                            let record_name: Name = Name::from_str_relaxed(name.as_str()).unwrap();
src/server/serve.rs:                let mut switch = kill_switch.lock().unwrap();
src/server/serve.rs:                let _ = handle.join().unwrap();
src/server/serve.rs:    tx.broadcast(true).await.unwrap();
src/server/serve.rs:    let address = address_string.parse().unwrap();
src/server/serve.rs:    let conn = UdpClientConnection::with_timeout(address, Duration::from_millis(5)).unwrap();
src/server/serve.rs:    let name = Name::from_str("anything.").unwrap();
```

We should always log a useful error and only exit when absolutely necessary. Aardvark is a daemon we should never die because of a small error.

flouthoc:
@Luap99 Above PR should close this issue.

PR #76, opened 2022-02-16
server, config: remove all remaining unwraps

flouthoc:
Removes some last remaining bits of unwrap usage

Edit: Although all the necessary unwrap were removed before but this makes sure we don't have any unwrap traces at all.

This completely closes the issue: https://github.com/containers/aardvark-dns/issues/59

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -271,17 +271,22 @@ async fn start_dns_server(
 
 #[tokio::main]
 async fn send_broadcast(tx: &async_broadcast::Sender<bool>) {
-    tx.broadcast(true).await.unwrap();
+    if let Err(e) = tx.broadcast(true).await {
+        error!("unable to broadcast to child threads: {:?}", e);
+    }
 }
 
 fn server_refresh_request(address_string: String) {
-    let address = address_string.parse().unwrap();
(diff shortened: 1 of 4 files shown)


src/server/serve.rs
```diff
@@ -271,17 +271,22 @@ async fn start_dns_server(
...
+        if let Ok(conn) = UdpClientConnection::with_timeout(address, Duration::from_millis(5)) {
+            // and then create the Client
+            let client = SyncClient::new(conn);
+            // server will be killed by last request
+            if let Ok(name) = Name::from_str("anything.") {
+                match client.query(&name, DNSClass::IN, RecordType::A) {
+                    _ => {}
+                }
+            }
+        }
```
Luap99, line 290: I do not understand why we need this? I think send_broadcast will already notify the listener threads

flouthoc: @Luap99 Yes this should not be needed anymore, I already added a marker here in the past: https://github.com/containers/aardvark-dns/blob/main/src/server/serve.rs#L191 but just want to test and do that in a different PR so its easier to review and revert.

flouthoc:
@Luap99 @baude Good to go. I think

Luap99:
I will take a look at the slirp timeout flake later

PR #81, opened 2022-02-17
tests: fix slirp flake

Luap99:
When we create a new netns with unshare we have to wait until it
completes before using the ns. When we run things in the background with
`&` we always need to check that the process is ready before using it.
We now have ready checks for unshare and slirp4netns so it should no
longer flake even on slow systems.

Also only setup slirp4netns when needed for the test to safe some setup
time.

### DIFF
--- a/test/100-basic-name-resolution.bats
+++ b/test/100-basic-name-resolution.bats
@@ -6,6 +6,8 @@
 load helpers
 
 @test "basic container - dns itself" {
+	setup_slirp4netns
+
 	subnet_a=$(random_subnet 5)
 	create_config "podman1" $(random_string 64) "aone" "$subnet_a" "a1" "1a"
 	config_a1=$config
--- a/test/200-two-networks.bats
+++ b/test/200-two-networks.bats
@@ -7,6 +7,8 @@ load helpers
 
 
 @test "two containers on different networks" {
+	setup_slirp4netns
+
 	# container a1 on subnet a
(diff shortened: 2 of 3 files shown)


Luap99:
I am unable to reproduce the flake locally, in a CI VM via hack/get_ci_vm.sh and here on the PR. I do not think this is a slirp4netns issue at this point. It works for me without any extra sleeps.

I think it is more likely environment related, it is possible that the upstream dns query's are dropped somewhere.

@baude @mheon @flouthoc WDYT?

baude:
likely it is the fact that the ci is resource constrained.  also, they running in parallel.  any chance you are getting errors caused by collisions?

baude:
has anyone tried with reproducing with manually with an actual ci instance? if not, maybe i can try.

Luap99:
> I am unable to reproduce the flake locally, in a CI VM via hack/get_ci_vm.sh and here on the PR. I do not think this is a slirp4netns issue at this point. It works for me without any extra sleeps.

@baude yes see above. I am sure that this is not a timing problem with slirp

Luap99:
Ok I think I found the root cause now. The problem is that the unshare call to create the netns is not fast enough. When we take the pid it is still in the parent ns before it calls setns so when slirp4netns read the netns from the pid it still has the old ns there. To fix this we need to wait until unshare created the netns.

Luap99:
This should be ready to merge now.

PR #85, opened 2022-02-21
format bats test files

Luap99:
Run a formatter against the bats files, there is no problem with the
current version but I think this looks a bit better.


flouthoc (approved):
LGTM
/lgtm

PR #89, opened 2022-02-23
Remove vendor directory from upstream repo

baude:
We no longer include the vendored libraries as part of the upstream git
repository.  Make vendor will populate the libraries as needed.


mheon:
LGTM

flouthoc (approved):
LGTM

Finally we can use dependabot.

PR #102, opened 2022-03-08
Drop vergen dependency

### LINKED ISSUE #93: Building from the releases tab fails.
This is a copy of an issue from containers/netavark#247 because it also happens when building **aardvark-dns**.

## Quick summary
Building from archives provided in the [Releases Page](https://github.com/containers/aardvark-dns/releases) fails due to lack of repository related information.
More Specifically vergen fails to embed git commit checksum since it doesn't exist.

### Steps to reproduce
1. download one of the *Source code* archives from the [Releases Page](https://github.com/containers/aardvark-dns/releases).
2. Extract the archive.
3. run `make` or `cargo build`.

### Error message
```
[...]
   Compiling git2 v0.13.25
   Compiling aardvark-dns v1.0.1 (/aardvark-dns-1.0.1)
error: failed to run custom build command for `aardvark-dns v1.0.1 (/aardvark-dns-1.0.1)`

Caused by:
  process didn't exit successfully: `/aardvark-dns-1.0.1/targets/release/build/aardvark-dns-55f410ec0c6374ff/build-script-build` (exit status: 1)
  --- stderr
  Error: could not find repository from '/aardvark-dns-1.0.1'; class=Repository (6); code=NotFound (-3)
make: *** [Makefile:47: build] Error 101
```

## Temporary solution
by removing "git" from list of vergen features in `Cargo.toml` and manually setting `VERGEN_GIT_SHA` environment variable I was able to build it successfully.
```shell
$ sed -i 's/, "git"//' Cargo.toml
$ env VERGEN_GIT_SHA="" make
```
Result:
```shell
$ ./bin/aardvark-dns version
{
  "version": "1.0.1",
  "commit": "",
  "build_time": "2022-02-25T21:29:25.854856292+00:00",
  "target": "x86_64-alpine-linux-musl"
}
```

Luap99:
We only need the version, build time, rust target triple, and git commit
for the version command. These values can easily be generated with little
code so we do not need an extra dependency for this.

The main reason for this change is that vergen will fail if no git
commit is found. On distro build systems this is often the case since
they build from a source tar without the git repo. Instead of erroring
we should just show an empty commit in the version output.

Fixes #93


baude:
code LGTM

PR #106, opened 2022-03-18
[v1.0.1-rhel] Backport "Drop vergen dependency"

Luap99:
Backport commit e824402c85d7.

We only need the version, build time, rust target triple, and git commit
for the version command. These values can easily be generated with little
code so we do not need an extra dependency for this.

The main reason for this change is that vergen will fail if no git
commit is found. On distro build systems this is often the case since
they build from a source tar without the git repo. Instead of erroring
we should just show an empty commit in the version output.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@
 # Main collection of env. vars to set for all tasks and scripts.
 env:
     # Actual|intended branch for this run
-    DEST_BRANCH: "main"
+    DEST_BRANCH: "v1.0.1-rhel"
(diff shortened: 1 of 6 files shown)


flouthoc (approved):
LGTM
/lgtm

TomSweeneyRedHat:
Hold this until approval from RHEL, then we'll need to create a new release.
But then again, I'm not seeing this vendored into the Podman v4.0-rhel branch?  How does that work with Rust into Golang?

mheon:
Aardvark is a separate project, Podman is not involved

PR #112, opened 2022-03-31
slirp4netns, test: increase max timeout limit for slirp4netns to start start

flouthoc:
Slirp4netns starts in uncertain way in different environment, increasing
timeout to ensure that we reduce flakes as much as possible. Since its
hard to reproduce this flake in different environment.

Last flake seen: https://github.com/containers/aardvark-dns/pull/111

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -503,7 +503,10 @@ function setup_slirp4netns() {
     echo "nameserver 10.0.2.3" >"$AARDVARK_TMPDIR/resolv.conf"
     run_in_host_netns mount --bind "$AARDVARK_TMPDIR/resolv.conf" /etc/resolv.conf
 
-    local timeout=6
+    # waiting for slirp4netns to start is uncertain in different environments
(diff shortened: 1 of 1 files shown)


flouthoc:
@Luap99 I am unable to reproduce this flake in different envs, slirp startup happens in really uncertain way in different envs. Saw a recent flake here: https://github.com/containers/aardvark-dns/pull/111

Luap99:
I am confident that this will not fix the flake, the timeout is already high enough. There is something else going on here.

flouthoc:
@Luap99 Its weird when i try this on local slirp starts without any issues on CI it flakes.

flouthoc:
I'm closing this PR will re-open if I fine any better fix, increasing timeout indeed does not fix the flake.

Luap99:
Because CI is super slow compared to anything you run local.
I think I found the problem, I can will open a PR later.

Luap99:
https://github.com/containers/aardvark-dns/pull/113

PR #113, opened 2022-03-31
tests: fix slirp flake v2

### LINKED ISSUE #110: fix the slirp related flake in integration tests
title says it all. we need to fix the flake.

Luap99:
Commit 4a451679ebbe points at the correct error source however the fix
for it was not working. ls -l will always look different when called on
different paths. We only want to compare the actual link target. Fix
this by using readlink.

Fixes #110

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -296,8 +296,8 @@ function create_netns() {
 
     # we have to wait for unshare and check that we have a new ns before returning
     local timeout=2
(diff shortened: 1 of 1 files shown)


flouthoc:
I think now integration is just stuck somewhere.

Luap99:
Should be working now.

flouthoc (approved):
/lgtm
/approve

PR #117, opened 2022-04-04
server, performance: no need to notify individual servers by making extra `UDP` connection.

flouthoc:
We are wasting CPU cycles while notifying servers whenever aardvark
performs a refresh, we don't need to do that anymore after we switched
to mpmc ( multi-producer, multi-consumer ) design for notification here
: https://github.com/containers/aardvark-dns/pull/48


Luap99 (commented):
LGTM

PR #137, opened 2022-05-09
task: add `tarpaulin` to check `code coverage` by unit tests

flouthoc:
Add task for running `tarpaulin` via `make` and in CI to generate code
coverage report by running unit test across the code

[NO TESTS NEEDED]
[NO NEW TESTS NEEDED]

Sample output after adding unit tests.

```console
May 09 12:15:35.108  INFO cargo_tarpaulin::report: Coverage Results:
|| Uncovered Lines:
|| src/backend/mod.rs: 64, 73, 81, 101, 110
|| src/commands/run.rs: 12, 16, 22, 27-30, 33, 38-39
|| src/commands/version.rs: 18-19, 33, 40, 42
|| src/config/mod.rs: 33-35, 73-75, 78-80, 90-91, 93, 144, 160-162, 202-204, 206-208, 212, 234-236, 238, 249-251, 263-265, 277-279, 281, 297-299, 301
|| src/dns/coredns.rs: 340-344, 346, 348, 350-351, 355, 358-362, 364, 366-367, 369-371, 373, 376, 379, 381, 383-385, 390-392, 394-398, 401-404, 407, 409-411
|| src/main.rs: 35, 39, 43-45, 47-48, 51, 54-57, 59, 62-63, 67, 69-71, 73-76, 79, 81-83
|| src/server/serve.rs: 27, 34-39, 42, 45-49, 55, 58-61, 67, 72, 74-76, 82, 85, 87-89, 91-93, 96, 101, 104-106, 109, 111-126, 128-131, 135, 138, 142-157, 159-161, 165, 168, 172-174, 178-181, 185-187, 192-193, 195, 197-200, 207, 216-221, 223-226, 230-235, 239-242, 249-251
|| src/test/test.rs: 18, 30, 37, 58, 61, 72, 75, 92, 95, 112, 114, 136, 139, 161, 164, 186, 189, 211, 214, 236, 239, 266, 269, 290, 293, 321, 375, 395
|| Tested/Total Lines:
|| src/backend/mod.rs: 25/30 +0.00%
|| src/commands/run.rs: 0/10 +0.00%
|| src/commands/version.rs: 0/5 +0.00%
|| src/config/mod.rs: 98/139 +0.00%
|| src/dns/coredns.rs: 0/44 +0.00%
|| src/main.rs: 0/27 +0.00%
|| src/server/serve.rs: 0/121 +0.00%
|| src/test/test.rs: 166/194 +0.00%
|| 
50.70% coverage, 289/570 lines covered, +10.394920355281513% change in coverage

```


flouthoc:
Could be merged before: https://github.com/containers/aardvark-dns/pull/138

Luap99:
IMO there is not much point in running the coverage job in CI unless we make the report visible with something like https://about.codecov.io/

Also the task already takes 13 min but it did not run any tests. This is only the time it took to download and compile tarpaulin. Given it takes that long we should move this directly into the VM images and not waste so much time on each PR.

flouthoc:
@Luap99 Actually I checked codecov but codecov itself uses `tarpaulin` and other tools but I think it does reports well on GUI . Output of `tarpaulin` is visible in task log of CI and can be also be generated on local.

> Also is it possible to get coverage data from the integration tests as well?

I think it should but not sure if it is generally acceptable and not in the way we run integration tests but anyways in general code-coverage is metric of line of code being covered by tests so `unit test` is a more natural approach to get code coverage i think this post here explains it better: https://sqa.stackexchange.com/a/25007
I think this is the reason that it becomes hard after a point to increase coverage of codebase cause some components can be only tested well via `integration test` but some of them can still be tested with `unit test`. Best way to increase coverage with time would be split components into small functions and write `unit test` for them.

Luap99:
> Output of `tarpaulin` is visible in task log of CI and can be also be generated on local.

Nobody will look at this until the job fails.

The code coverage tools can show you which lines are not covered when you push new code which will be helpful for reviewers to see if the code is actually tested.

flouthoc:
Let me see if there is a way to generate gui on top of this or use `codecov` with `tarpaulin`

flouthoc:
Yes embeding this into `VM` should help although I think we can drop this from CI and just let it be on `makefile` if we enable `codecov` then this is just wasting compute in CI.

flouthoc:
> Yes embeding this into `VM` should help although I think we can drop this from CI and just let it be on `makefile` if we enable `codecov` then this is just wasting compute in CI.

Dropped this from CI and only keeping in `make` targets so we save compute and let `codecov` do the thing cause it uses the same tool under the hood.

Luap99 (commented):
LGTM

flouthoc:
@mheon @baude @lsm5 PTAL

PR #141, opened 2022-05-13
Cirrus: enable Fedora 36 aarch64 tasks on EC2

lsm5:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,24 +26,18 @@ env:
 
 gcp_credentials: ENCRYPTED[f6a0e4101418bec8180783b208721fc990772817364fed0346f5fd126bf0cfca03738dd8c7fb867944637a1eac7cec37]
 
-
-# Default VM to use unless set or modified by task
-gce_instance: &standard_vm
-    image_project: "libpod-218412"
-    zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
-    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
-    image_name: "${FEDORA_NETAVARK_IMAGE}"
-
+aws_credentials: ENCRYPTED[3fab904a98355f84b0bac084f8a50428ff8a27dd2b6a6c42fca77df89010e620a1da3cd246a50b1074e6787c42818080]
 
 build_task:
   alias: "build"
   # Compiling is very CPU intensive, make it chooch quicker for this task only
-  gce_instance:
-    <<: *standard_vm
+  gce_instance: &standard_build_gce_x86_64
+    image_project: "libpod-218412"
+    zone: "us-central1-c"
+    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
     cpu: 8
     memory: "8Gb"
+    image_name: "${FEDORA_NETAVARK_IMAGE}"
   cargo_cache: &cargo_cache
     folder: "$CARGO_HOME"
     fingerprint_key: "cargo_v1_${DEST_BRANCH}"  # Cache for branch + PRs
@@ -62,10 +56,36 @@ build_task:
   upload_caches: [ "cargo", "targets", "bin" ]
 
 
+build_aarch64_task:
+  alias: "build_aarch64"
+  # Compiling is very CPU intensive, make it chooch quicker for this task only
+  ec2_instance: &standard_build_ec2_aarch64
+    image: ami-0f89dfa699fe590b0
+    type: t4g.xlarge
+    region: us-east-1
+    architecture: arm64  # CAUTION: This has to be "arm64", not aarch64.
+  cargo_cache: &cargo_cache_aarch64
+    <<: *cargo_cache
+    fingerprint_key: "cargo_v1_${DEST_BRANCH}_aarch64"
+  targets_cache: &targets_cache_aarch64
+    <<: *targets_cache
+    fingerprint_key: "cargo_v1_${DEST_BRANCH}_aarch64"
+  bin_cache: &bin_cache_aarch64
+    <<: *bin_cache
+    fingerprint_key: "cargo_v1_${DEST_BRANCH}_aarch64"
+  setup_script: *setup
+  main_script: *main
+  upload_caches: [ "cargo", "targets", "bin" ]
+
+
 validate_task:
   alias: "validate"
   depends_on:
     - "build"
+  gce_instance: &standard_gce_x86_64
+    <<: *standard_build_gce_x86_64
+    cpu: 2
+    memory: "4Gb"
   # From this point forward, all cache's become read-only for this run.
   cargo_cache: &ro_cargo_cache
     <<: *cargo_cache
@@ -80,13 +100,45 @@ validate_task:
   main_script: *main
 
 
+validate_aarch64_task:
+  alias: "validate_aarch64"
+  depends_on:
+    - "build_aarch64"
+  ec2_instance: *standard_build_ec2_aarch64
+  # From this point forward, all cache's become read-only for this run.
+  cargo_cache: &ro_cargo_cache_aarch64
+    <<: *cargo_cache_aarch64
+    reupload_on_changes: false
+  targets_cache: &ro_targets_cache_aarch64
+    <<: *targets_cache_aarch64
+    reupload_on_changes: false
+  bin_cache: &ro_bin_cache_aarch64
+    <<: *bin_cache_aarch64
+    reupload_on_changes: false
+  setup_script: *setup
+  main_script: *main
+
+
 verify_vendor_task:
   alias: "verify_vendor"
   depends_on:
     - "build"
+  gce_instance: *standard_gce_x86_64
   cargo_cache: *ro_cargo_cache
   targets_cache: *ro_targets_cache
   bin_cache: *ro_bin_cache
+  setup_script: *setup
+  main_script: *main
+
+
+verify_vendor_aarch64_task:
+  alias: "verify_vendor_aarch64"
+  depends_on:
+    - "build_aarch64"
+  ec2_instance: *standard_build_ec2_aarch64
+  cargo_cache: *ro_cargo_cache_aarch64
+  targets_cache: *ro_targets_cache_aarch64
+  bin_cache: *ro_bin_cache_aarch64
   setup_script: *setup
   main_script: *main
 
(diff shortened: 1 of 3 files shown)


lsm5:
hmm, looks like copying the image id from fedora website as-is won't work and we need fedora's magic button powers, or perhaps a custom ami built on fedora's default image.

lsm5:
alright, so before I can use Amazon's own aarch64 image, I'll need to accept their ToS. I'll need to look further if it adds any pointless costs, if it does not, that might be a convenient way to proceed.

/cc @cevich

.cirrus.yml
```diff
@@ -167,6 +167,17 @@ centos9_build_task:
     script:
         - cargo build
 
+aws_credentials: ENCRYPTED[3fab904a98355f84b0bac084f8a50428ff8a27dd2b6a6c42fca77df89010e620a1da3cd246a50b1074e6787c42818080]
+
+#image: ami-0d15f8a570d61e229
+#image: ami-01925eb0821988986
+fedora36_aarch64_task:
+  image: ami-0b8b7e5c3c2df76c7
```
cevich, line 175: This line needs to be nested inside the `ec2_instance` map down below :grin:

cevich: See example here: https://cirrus-ci.org/guide/supported-computing-services/#ec2

lsm5: done

cevich:
@fkorotkov FYI - this is the PR Lokesh is trying to get working in AWS on aarch64.

@lsm5 I sent you an e-mail, there's a good chance that my meddling with the IAM policy caused the `You are not authorized to perform this operation.`  I'll see if I can track it down in the logs and report back.

cevich:
...yep, it's my fault.  I made our cirrus EC2 policy deny access to AMI's we don't own. I did this not only for slightly tighter security -  Many public images (in both AWS and GCP) have defined "expiration" dates and/or get depreciated on the whim of whomever/whatever maintains them.  So this causes surprises, especially for tests that end up in long-term release branches, as they will eventually, unexpectedly, and irrecoverably break at some future date.

So long-term, eventually you'll want to use AMI's produced by the `c/automation_images` build workflow (where we remove the image expiration date).  But for now, let me see if I can manually copy that Fedora project produced AMI to a private one for you...

cevich:
@lsm5 okay, try `ami-087583f5b71595ad8` (copy of `ami-01925eb0821988986`).

fkorotkov:
@cevich it seems logs are not saved in your S3. Please make sure you've add permissions to access S3 like mentioned in the [docs](https://cirrus-ci.org/guide/supported-computing-services/#aws).

cevich:
Oof, yeah that's probably my fault too.   I made a policy to restrict cirrus-ci users to specifically named buckets.  I thought I fixed it for this user but apparently not, and the logs don't say why :cry:

@lsm5 can you try re-pushing to this PR instead of re-running, maybe a fresh build will reset state and get the logs flowing properly.

lsm5:
> Oof, yeah that's probably my fault too. I made a policy to restrict cirrus-ci users to specifically named buckets. I thought I fixed it for this user but apparently not, and the logs don't say why cry
> 
> @lsm5 can you try re-pushing to this PR instead of re-running, maybe a fresh build will reset state and get the logs flowing properly.

worked. Can see the log now.

cevich:
> worked. Can see the log now.

Great, so @fkorotkov I'm observing that initial access to the storage bucket is established per-build, not per task.  That state seems to persist, so if a task is re-run but bucket access was originally denied to the build, logging will fail for the re-run.  This isn't a problem per-say, just something to be aware of for us kids who might go tinkering with the recommended IAM policy :stuck_out_tongue_winking_eye:

cevich:
@lsm5 finally [got a build to complete](https://cirrus-ci.com/task/5739723703451648).  Give `ami-0ef09d8442b7f877e` a try here.  It's got all the standard tooling and stuff installed as the GCE VM image for netavark/aardvark CI (but arm64 of course).

contrib/cirrus/setup.sh
```diff
@@ -26,4 +26,7 @@ curl --fail --location -o /tmp/netavark.zip "$NETAVARK_URL"
 mkdir -p /usr/libexec/podman
 cd /usr/libexec/podman
 unzip /tmp/netavark.zip
+if [[ $(uname -m) != "x86_64" ]]; then
```
cevich, line 29: May need a `rm -f netavark` before the unzip line.   Or tell unzip to overwrite everything.

lsm5:
@baude @flouthoc @mheon @Luap99 can I get some eyes on `integration_aarch64` please?

flouthoc:
From logs it seems all of the tests failed which makes me think that aardvark or netavark is not running at all. Could you make sure if binary is of the correct arch ? running `./aardvark-dns --version` or `./aardvark-dns --help` on the server should confirm if the binary is correct. Issue could be also with the `netavark` binary itself so it must be checked as well.

One issue i also see is with `unit` tests if you see there are logs of `unit` tests but no logs for `unit_aarch64` is that expected.

lsm5:
alright everything passes now.

I suspect the last run was an issue `integration_aarch64` not being able to see the logs of the previous runs.

@cevich I don't know what happened but the S3 permissions you had set last week suddenly stopped working this week. I had to adjust the S3 permissions further to make the logs viewable.

lsm5:
This should be good to go.

@cevich please lgtm when you're back :)

.cirrus.yml
```diff
@@ -62,10 +56,30 @@ build_task:
...
 
+build_aarch64_task:
+  alias: "build_aarch64"
+  # Compiling is very CPU intensive, make it chooch quicker for this task only
+  ec2_instance: &standard_build_ec2_aarch64
+    image: ami-0f89dfa699fe590b0
+    type: t4g.xlarge
+    region: us-east-1
+    architecture: arm64  # CAUTION: This has to be "arm64", not aarch64.
+  cargo_cache: *cargo_cache
```
cevich, line 67: This task is going to cause cache-clashes with the build task.  This will cause a race in dependent tasks over which cache to use (since the key is the same).  So we need a unique cache key for the non-x86_64 tasks to prevent this.

```suggestion
  cargo_cache: &cargo_cache_aarch64
      <<: *cargo_cache
      fingerprint_key: "cargo_v1_${DEST_BRANCH}_aarch64"
```

cevich: Note: This applies to ALL the `aarch64` cache-related anchors and aliases :cry:

.cirrus.yml
```diff
@@ -80,10 +94,36 @@ validate_task:
...
 
+validate_aarch64_task:
+  alias: "validate_aarch64"
+  depends_on:
+    - "build_aarch64"
+  ec2_instance: *standard_build_ec2_aarch64
+  # From this point forward, all cache's become read-only for this run.
+  cargo_cache: *ro_cargo_cache
+  targets_cache: *ro_targets_cache
+  bin_cache: *ro_bin_cache
```
cevich, line 105: All three of these need to be updated to use distinct `*_aarch64` cache IDs (aka `fingerprint_key`).
```suggestion
  bin_cache: &ro_bin_cache_aarch64
      <<: *ro_bin_cache
      fingerprint_key: "bin_v1_${DEST_BRANCH}_aarch64"
```

.cirrus.yml
```diff
@@ -80,10 +94,36 @@ validate_task:
...
+  setup_script: *setup
+  main_script: *main
+
+
+verify_vendor_aarch64_task:
+  alias: "verify_vendor_aarch64"
+  depends_on:
+    - "build_aarch64"
+  ec2_instance: *standard_build_ec2_aarch64
```
cevich, line 126: The cache lines below should alias the `_aarch64` anchors, for example `cargo_cache: *ro_cargo_cache_aarch64`

.cirrus.yml
```diff
@@ -95,6 +135,19 @@ unit_task:
...
+  bin_cache: *ro_bin_cache
+  setup_script: *setup
+  main_script: *main
+
+
+unit_aarch64_task:
+  alias: "unit_aarch64"
+  depends_on:
+    - "build_aarch64"  # Run in parallel with validate to save some time
+  ec2_instance: *standard_build_ec2_aarch64
```
cevich, line 150: Same for the `_cache` entries here.

.cirrus.yml
```diff
@@ -106,6 +159,19 @@ integration_task:
...
+  bin_cache: *ro_bin_cache
+  setup_script: *setup
+  main_script: *main
+
+
+integration_aarch64_task:
+  alias: "integration_aarch64"
+  depends_on:
+    - "unit_aarch64"
+  ec2_instance: *standard_build_ec2_aarch64
```
cevich, line 174: Here too

cevich (changes requested):
In general this is basically read to go.  I posted a handful of comments for changes to the `fingerprint_key` for all aarch64 tasks cache.  If not corrected, and some future PR adds non-aarch64 tasks to AWS, there will be a very non-obvious race-condition / cache-id clash.  This should be accounted for by using unique cache keys per architecture.

lsm5:
@cevich i think i have addressed all your comments on this PR. But for whatever reason, the `integration_aarch64` task fails most likely because it can't find an aardvark-dns binary. AFAICT, it looks for binary at `/var/tmp/aardvark-dns/test/../bin/aardvark-dns` but when I re-run with Terminal Access, there's no binary to be found there. After manually running make, make integration passes.

cevich:
> After manually running make, make integration passes.

Ahh, this is the key clue.  So something's breaking with cache then, I'l take a peek.

.cirrus.yml
```diff
@@ -138,6 +216,19 @@ fedora36_build_task:
...
       - "build"
+    gce_instance: *standard_gce_x86_64
+    container:
+        cpu: 2
+        memory: 2
+        image: quay.io/libpod/fedora36rust
+    script:
+        - cargo build
+
+fedora36_build_aarch64_task:
```
cevich, line 227: Sorry didn't notice this before.  This task is redundant w/ `build_aarch64_task` (also F36, prev. was F35).

cevich:
> most likely because it can't find an aardvark-dns binary

I looked over the changes and the existing scripts and I think this might be a test problem.  The key here is (presumably) this is not a problem for the x86_64 tests, so that makes me think it's an environment problem.  A few suggestions of things to confirm:

1. Make sure in all cases, neither the netavark nor the aardvark-dns RPMs are installed on the VM.  If so, this is an image-build bug.
2. Check if maybe the tests are sensitive to some env. var. to indicate where the test-binary is located.
3. Check if the tests are sensitive to executing from a certain directory, maybe in the aarch64 case $CWD is something different and the tests expect a hard-coded `./bin/aardvark-dns`.

lsm5:
> Hrmmm, not sure why it's failing. Is this the same error?

you mean the `integration` task for x86_64? seems to pass now.

PR #143, opened 2022-05-23
Cirrus: Use fixed netavark branch for CI

cevich:
This is important for the stability of CI in case of a future backport
that happens to be incompatible with netavark `main`.  Since CI
doesn't run very often on the aardvark-dns `v1.0.1-rhel` branch, an
incompatible change may not be noticed.  Fix this by switching off
of the `main` branch onto a netavark release branch.

Signed-off-by: Chris Evich <cevich@redhat.com>


flouthoc (approved):
LGTM

PR #146, opened 2022-06-08
lookup: make name lookup case insensitive as per `RFC4343`

flouthoc:
As per https://datatracker.ietf.org/doc/html/rfc4343 DNS name resolution
must be case insensitive so convert everything to lowercase by
performing a lookup and storing names into the backend.

Closes: https://github.com/containers/podman/issues/14525


baude:
LGTM

mheon:
LGTM

PR #153, opened 2022-06-28
[DO NOT MERGE] Check if build job works

lsm5:
[NO NEW TESTS NEEDED]

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -24,19 +24,6 @@ env:
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
 
 
-gcp_credentials: ENCRYPTED[f6a0e4101418bec8180783b208721fc990772817364fed0346f5fd126bf0cfca03738dd8c7fb867944637a1eac7cec37]
-
-
-# Default VM to use unless set or modified by task
-gce_instance: &standard_vm
-    image_project: "libpod-218412"
-    zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
-    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
-    image_name: "${FEDORA_NETAVARK_IMAGE}"
-
-
 cargo_cache_common: &cargo_cache
     folder: "$CARGO_HOME"
     fingerprint_key: "cargo_v1_${DEST_BRANCH}"  # Cache for branch + PRs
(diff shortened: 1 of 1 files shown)


lsm5:
please don't close this yet.

.cirrus.yml
```diff
@@ -37,28 +37,60 @@ gce_instance: &standard_vm
...
+aws_credentials: ENCRYPTED[3fab904a98355f84b0bac084f8a50428ff8a27dd2b6a6c42fca77df89010e620a1da3cd246a50b1074e6787c42818080]
+
+# Default aarch64 AWS instance to use unless set or modified by task
+ec2_instance: &standard_aarch64_instance
+  image: ami-0a72cf7a8b5b513d4
+  type: t4g.micro
+  region: us-east-1
+  architecture: arm64
+
+build_aarch64_task:
```
lsm5, line 85: @cevich trying to add aarch64 tasks like so ^. The aarch64 task succeeds but the regular `build` task seems to wait forever. If I comment out the aarch64 task, the `build` task works as expected. Any idea why that happens?

contrib/cirrus/setup.sh
```diff
@@ -25,5 +25,5 @@ set -x  # show what's happening
 curl --fail --location -o /tmp/netavark.zip "$NETAVARK_URL"
 mkdir -p /usr/libexec/podman
 cd /usr/libexec/podman
-unzip /tmp/netavark.zip
-chmod a+x /usr/libexec/podman/netavark
+#unzip -q /tmp/netavark.zip
+#chmod a+x /usr/libexec/podman/netavark
```
lsm5, line 29: For aarch64, unzipping netavark.zip errored out on waiting for `replace netavark`. So, I commented this out.

lsm5:
@baude @cevich so if I only keep ec2 tasks and no gce tasks, then CI is all green. So, it looks like only with ec2 and gce enabled together, the gce `build` task consistently times out after a long time.

I'll try spawning ec2 machines for x86_64 (maybe in another PR).

contrib/cirrus/setup.sh
cevich: Right, so the problem with this is there is no arm64 build to restore, only x86, so that ain't gonna work.  Let me take a peek and refresh my memory how this all works.

.cirrus.yml
```diff
@@ -24,48 +24,52 @@ env:
...
-
-
-# Default VM to use unless set or modified by task
-gce_instance: &standard_vm
-    image_project: "libpod-218412"
-    zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
-    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
-    image_name: "${FEDORA_NETAVARK_IMAGE}"
```
cevich, line 37: Yeah, I think next step is to remove this, and the top-level `ec2_instance` block.  Define them in separate tasks, and set the YAML anchors there for re-use (if needed).  The top-level definitions are suppose to be "defaults" for when a task doesn't define one.  I have a strong feeling that having TWO defaults is going to really confuse Cirrus-CI.

.cirrus.yml
```diff
@@ -24,48 +24,52 @@ env:
...
     folder: "$CIRRUS_WORKING_DIR/bin"
     fingerprint_key: "bin_v1_${CIRRUS_BUILD_ID}" # Cache only within same build
     reupload_on_changes: true
-  setup_script: &setup "$SCRIPT_BASE/setup.sh $CIRRUS_TASK_NAME"
-  main_script: &main "$SCRIPT_BASE/runner.sh $CIRRUS_TASK_NAME"
+
+
+setup_script: &setup "$SCRIPT_BASE/setup.sh $CIRRUS_TASK_NAME"
+
+main_script: &main "$SCRIPT_BASE/runner.sh $CIRRUS_TASK_NAME"
```
cevich, line 46: The Cirrus-CI parser may be forgiving of this, but I wouldn't recommend it.  Their docs specify a schema for the top-level block names (basically, `*_instance`, `*_credentials`, `env`, and `*_task`.  It's probably best to stick to just those, less some future update to their parser suddenly break this non-conforming usage.

PR #156, opened 2022-06-30
Move instead of borrow to avoid need to clone

j0057:
This gets rids of a small TODO issue, and avoids copying three potentially large HashMaps.


Luap99 (approved):

flouthoc:
Looks like delay in merge bot

j0057:
Thanks for merging!

PR #159, opened 2022-07-11
Makefile: add target for release tarball

Luap99:
Add a target to automatically create the vendor tarball and the extra
files. This can only be used on a version which has not the -dev suffix
to prevent user errors.


flouthoc (approved):
LGTM

flouthoc:
@mheon @baude PTAL

PR #170, opened 2022-07-18
Cirrus: Verify expected binary artifacts present

cevich:
Other CI systems depend/expect a full set of compiled binaries are
included in the success task's binary artifact (zip file).  However,
it's possible some mistake in CI could silently lead to one or more
binaries not being produced.  For example, a `bin_cache` failure.  Fix
this by verifying all expected/required files are present.  If not,
print the name of the missing binary to assist in
troubleshooting/debugging.

Signed-off-by: Chris Evich <cevich@redhat.com>

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -79,10 +79,12 @@ build_aarch64_task:
   # However, we don't want to confuse architectures.
   art_prep_script:
     - cd bin
-    - ln aardvark-dns aardvark-dns.$(uname -m)-unknown-linux-gnu
-    - ln aardvark-dns.debug aardvark-dns.debug.$(uname -m)-unknown-linux-gnu
+    - ls -la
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -273,6 +283,10 @@ success_task:
     - unzip /tmp/armbinary.zip
     - mv bin/* ./
     - rm -rf bin
+  artifacts_test_script:  # Other CI systems depend on all files being present
+    - ls -la
```
flouthoc, line 287: What does this `ls -la` is this for debugging.

cevich: Yes it shows the files that are actually present, since they are renamed in the build task.  It makes it easier to diagnose any test failure, as opposed to requiring a developer to manually download the artifact zip to check.

flouthoc: SGTM thanks for the explanation :)

flouthoc (approved):
LGTM

PR #183, opened 2022-07-29
config: use multiple ips per line

Luap99:
We have to support several ipv4 and v6 addresses per container. The
current way of creating the configs in aardvark are just broken IMO. We
should never have to duplicate the container id and aliases just to set
more ips.

This patch is fully backwards compatible and will allow much better
config generation in netavark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,5 @@ make
 ```console
 RUST_LOG=trace ./bin/aardvark-dns --config src/test/config/podman/ --port 5533 run
 ```
+
+### [Configuration file format](./config.md)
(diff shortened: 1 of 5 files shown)


Luap99:
@flouthoc @mheon PTAL

mheon:
LGTM

flouthoc (changes requested):
LGTM, @Luap99 Could you add a section to Readme.md or probably a new `config.md` in this repo, describing the format supported by config. Since as of now we don't have any public format documented. ( Not a blocker )

Luap99:
> LGTM, @Luap99 Could you add a section to Readme.md or probably a new `config.md` in this repo, describing the format supported by config. Since as of now we don't have any public format documented. ( Not a blocker )

done

PR #190, opened 2022-08-08
Fix makefile comparisons

hoshsadiq:
This fixes the usage of `[[ ... ]]` since it does not work with `sh`, which `make` uses by default. Using `[ ... ]`/`case ... esac` works fine. 

These changes are similar to https://github.com/containers/netavark/pull/350.


flouthoc (approved):
LGTM
/lgtm
/approve

Issue #203, opened 2022-08-26
Always return both A and AAAA records no matter what QTYPE is specified in DNS request


xiaoyar:
# Version of aardvark-dns
```
Name        : aardvark-dns
Version     : 1.0.2
Release     : 1.el8
Architecture: x86_64
```

# Configuration
## Dual stack network
```
{
  "name": "dual",
  "id": "2697203bf4180da9e7a6d074e38cbafb2fad4c8a3436522bde4ac573c059caa6",
  "driver": "bridge",
  "network_interface": "podman1",
  "created": "2022-08-24T04:03:37.236675178-05:00",
  "subnets": [
    {
      "subnet": "192.168.227.0/24",
      "gateway": "192.168.227.1"
    },
    {
      "subnet": "fdf8:192:168:227::/120",
      "gateway": "fdf8:192:168:227::1"
    }
  ],
  "ipv6_enabled": true,
  "internal": false,
  "dns_enabled": true,
  "ipam_options": {
    "driver": "host-local"
  }
}
```

## Two containers, foo and bar, both running Nginx, by following configuration, Nginx in foo forwards request to Nginx in bar.
```
        location /bar {
                resolver        192.168.227.1;
                set $upstream   bar.dns.podman;
                proxy_pass http://$upstream;
        }
```

# Issue observed
## The web request fails with 502 error.
```
[root@foo /]# curl -vvv http://localhost/bar
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 80 (#0)
> GET /bar HTTP/1.1
> Host: localhost
> User-Agent: curl/7.61.1
> Accept: */*
>
< HTTP/1.1 502 Bad Gateway
```

## It's because Nginx fails to resolve hostname of "bar.dns.podman".
We can see Nginx error.log is filled up with plenty of the following errors:
```
2022/08/26 09:54:58 [error] 88#0: unexpected AAAA record in DNS response
2022/08/26 09:54:58 [error] 88#0: unexpected A record in DNS response
```

## The root cause is because `aardvark-dns` always returns both A and AAAA records no matter what QTYPE is specified in DNS request.
```
[root@foo /]# nslookup -type=A bar 192.168.227.1
Server:		192.168.227.1
Address:	192.168.227.1#53

Non-authoritative answer:
Name:	bar.dns.podman
Address: 192.168.227.5
Name:	bar.dns.podman
Address: fdf8:192:168:227::5

[...]

PR #205, opened 2022-08-26
coredns: do not combine results of `A` and `AAAA` records when query type only requests specific result.

flouthoc:
As of now aardvark-dns combines the result of `A` and `AAAA` record even
when request type is `A` or `AAAA` lets not do that and only return
returns for the requested type.

Easily testable using
```
nslookup -type=A bar 192.168.227.1
or
nslookup -type=AAAA bar 192.168.227.1
```

Closes: https://github.com/containers/aardvark-dns/issues/203


flouthoc:
@xiaoyar If easily possible could you try the patch otherwise I'll just add equivalent test anyways.

flouthoc:
Tests must be retrofitted for this bug fix, i'll retrofit tests.

mheon:
LGTM once you get the tests fixed

flouthoc:
**Linked BZ : #2124414**

flouthoc:
Retrofitted the tests, I hope podman and netavark are happy with this as is.

test/helpers.bash
```diff
@@ -558,7 +558,11 @@ function dig() {
     # first arg is container_netns_pid
     # second arg is name
     # third arg is server addr
-    run_in_container_netns "$1" "dig" "+short" "$2" "@$3"
+    if [ ! -z "$4" ]; then
+        run_in_container_netns "$1" "dig" "+short" "$2" "@$3" "$4"
+    else
+        run_in_container_netns "$1" "dig" "+short" "$2" "@$3"
+    fi
```
Luap99, line 565: just leave the $4 unquoted then you do not need the if else, unless you ever expect that the type could contain special chars or spaces.

test/100-basic-name-resolution.bats
```diff
@@ -53,15 +53,15 @@ load helpers
...
 	# Set recursion bit is already set if requested so output must not
 	# contain unexpected warning.
 	assert "$output" !~ "WARNING: recursion requested but not available"
 
-	run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw"
-	# validate that we get an ipv4
-	assert "$output" =~ "[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+"
+	run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw" "AAAA"
+	# validate that we get an ipv6
+	assert "$output" =~ "[0-9a-fA-F]+\:[0-9a-fA-F]+\:[0-9a-fA-F]+\:[0-9a-fA-F]+"
```
Luap99, line 64: I am confused about this regex? It doesn't actually validate a proper ipv6. You could have an address like `fd11::1`

test/helpers.bash
flouthoc: Yeah sounds better.

test/100-basic-name-resolution.bats
Luap99: If we have ipcalc installed in the VMs we can use: `ipcalc -c $ip`

flouthoc: I think it should be installed by default, I can amend test and see if it passes otherwise i can add a `TODO` comment here and once automation image will get updated we can replace the test with `ipcalc`

test/100-basic-name-resolution.bats
```diff
@@ -53,15 +53,15 @@ load helpers
...
 	# Set recursion bit is already set if requested so output must not
 	# contain unexpected warning.
 	assert "$output" !~ "WARNING: recursion requested but not available"
 
-	run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw"
-	# validate that we get an ipv4
-	assert "$output" =~ "[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+"
+	run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw" "AAAA"
+	# validate that we got valid ipv6
+	expected_rc=0 run_helper ip_calc -c "$output"
```
Luap99, line 64: Why expected_rc=0? That should be the default. Also you have to use `ip_calc -6c` because otherwise it will validate both ipv4 and ipv6 addresses and this test must make sure it is an ipv6

flouthoc: My bad was not aware the ex_rc=0 is the default and converted to only ipv6.

Luap99 (commented):
LGTM

flouthoc:
CI says `ipcalc -6c 2607:f8b0:4001:c1f::71` is bad Ipv6 while same command on my local passes, this is strange.

Luap99:
> CI says `ipcalc -6c 2607:f8b0:4001:c1f::71` is bad Ipv6 while same command on my local passes, this is strange.

Because the output contains more than one ip address and we try to validate all at once.

Use something like this:
```
# check that the output is not empty
assert "$lines[0]" != "" "got at least one result"
for ip in "${lines[@]}"; do
	ipcal -6c "$ip"
done
```

flouthoc:
This is go to go all test passes for me from podman's main branch and tests in `test/e2e/network_test.go`

flouthoc:
If this goes green the PR is good to merge otherwise it might need https://github.com/containers/aardvark-dns/pull/230

flouthoc:
> This is go to go all test passes for me from podman's main branch and tests in `test/e2e/network_test.go`

@Luap99 could you lgtm again, please. Already tested this on my local with `podman` tests.

flouthoc:
@mheon @baude PTAL

flouthoc:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2124414

mheon:
There a reason this still has a hold on it?

flouthoc:
I forgot hold cancel
/hold cancel

Issue #204, opened 2022-08-26
Recursion Available bit is not set in response header


xiaoyar:
# Version of aardvark-dns
```
Name        : aardvark-dns
Version     : 1.0.2
Release     : 1.el8
Architecture: x86_64
```

# Configuration
## Dual stack network
```
{
  "name": "dual",
  "id": "2697203bf4180da9e7a6d074e38cbafb2fad4c8a3436522bde4ac573c059caa6",
  "driver": "bridge",
  "network_interface": "podman1",
  "created": "2022-08-24T04:03:37.236675178-05:00",
  "subnets": [
    {
      "subnet": "192.168.227.0/24",
      "gateway": "192.168.227.1"
    },
    {
      "subnet": "fdf8:192:168:227::/120",
      "gateway": "fdf8:192:168:227::1"
    }
  ],
  "ipv6_enabled": true,
  "internal": false,
  "dns_enabled": true,
  "ipam_options": {
    "driver": "host-local"
  }
}
```

## Two containers, foo and bar, resolving bar from foo
```
[root@foo /]# cat /etc/resolv.conf
search dns.podman
nameserver 192.168.227.1
nameserver fdf8:192:168:227::1
```

# Issue observed
## `nslookup` complains "Got recursion not available from 192.168.227.1, trying next server"
```
[root@foo /]# nslookup bar
;; Got recursion not available from 192.168.227.1, trying next server
;; connection timed out; no servers could be reached

[root@foo /]#
```

## `dig` also complains "WARNING: recursion requested but not available"
```
[root@foo /]# dig bar

; <<>> DiG 9.11.36-RedHat-9.11.36-3.el8 <<>> bar
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 13400
;; flags: qr rd ad; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: b8dbf9748e7ba467 (echoed)
;; QUESTION SECTION:
;bar.				IN	A

;; ANSWER SECTION:
bar.			86400	IN	A	192.168.227.9
bar.			86400	IN	AAAA	fdf8:192:168:227::9
bar.			86400	IN	A	192.168.227.9
bar.			86400	IN	AAAA	fdf8:192:168:227::9

[...]

PR #206, opened 2022-08-26
coredns: response message always set recursion available (`RA`) when `RD` is set in request

flouthoc:
Set recursion available in all response message by setting `RA` flag in
response packet.

Closes: https://github.com/containers/aardvark-dns/issues/204

### DIFF
--- a/test/100-basic-name-resolution.bats
+++ b/test/100-basic-name-resolution.bats
@@ -17,10 +17,16 @@ load helpers
 	a1_pid=$CONTAINER_NS_PID
 	run_in_container_netns "$a1_pid" "dig" "+short" "aone" "@$gw"
 	assert "$ip_a1"
+	# Set recursion bit is already set if requested so output must not
+	# contain unexpected warning.
+	assert "$output" !~ "WARNING: recursion requested but not available"
 
 	run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw"
 	# validate that we get an ipv4
 	assert "$output" =~ "[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+"
+	# Set recursion bit is already set if requested so output must not
+	# contain unexpected warning.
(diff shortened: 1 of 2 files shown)


flouthoc:
PR should be good but lets wait for https://github.com/containers/aardvark-dns/pull/208

flouthoc:
PR is good for review and merge. @mheon @baude PTAL

test/100-basic-name-resolution.bats
```diff
@@ -49,10 +53,14 @@ load helpers
 	a1_pid=$CONTAINER_NS_PID
 	run_in_container_netns "$a1_pid" "dig" "+short" "aone" "@$gw"
 	assert "$ip_a1"
+	# Set recursion bit
+        assert "$output" !~ "WARNING: recursion requested but not available"
```
mheon, line 57: Nit: spacing on these asserts is weird - doesn't match the rest of the file

flouthoc: My bad, i used spaces instead of tabs and added extra space.

mheon:
CI seems broken, but it doesn't look like your PR; could be a more general thing?

flouthoc:
@cevich Looks like the new VM is missing configured rust stable, could it happen cause we update the VM.

cevich:
> could it happen cause we update the VM.

Most probably.  Here's what happes at image-build time:

https://github.com/containers/automation_images/blob/main/cache_images/fedora-netavark_packaging.sh#L72-L76

So if I understand correctly, we need to add 'default' to the end of that?  Better yet, since y'all probably know better than I, just open up a PR in that repo. to fix it.  It will build new images in about an hour that you can then test in this PR.

flouthoc:
This is good to merge now.

flouthoc:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2124416

PR #214, opened 2022-08-31
[CI:BUILD] Packit: initial enablement

lsm5:
This commit will run Koji scratch builds on every PR against all active
releases of Fedora, thus allowing buildability checks before the
PR merges.

### DIFF
--- a/.packit.sh
+++ b/.packit.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This script handles any custom processing of the spec file generated using the `post-upstream-clone`
+# action and gets used by the fix-spec-file action in .packit.yaml.
+
+set -eo pipefail
+
+# Get Version from HEAD
+VERSION=$(grep '^version' Cargo.toml | cut -d\" -f2 | sed -e 's/-/~/')
+
+# Generate source tarball
+git archive --prefix=aardvark-dns-$VERSION/ -o aardvark-dns-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Fix Version
+sed -i "s/^Version:.*/Version: $VERSION/" aardvark-dns.spec
+
+# Fix Release
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" aardvark-dns.spec
+
+# Fix Source0
+sed -i "s/^Source:.*.tar.gz/Source: aardvark-dns-$VERSION.tar.gz/" aardvark-dns.spec
+
+# Fix autosetup
+sed -i "s/^%autosetup.*/%autosetup -Sgit -n %{name}-$VERSION/" aardvark-dns.spec
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,25 @@
+---
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+# Build targets can be found at:
+# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
+
+specfile_path: aardvark-dns.spec
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    owner: rhcontainerbot
+    project: packit-builds
+    enable_net: true
+    srpm_build_deps:
+      - cargo
+      - make
+      - openssl-devel
+      - rpkg
+    actions:
+      post-upstream-clone:
+        - "rpkg spec --outdir ./"
+      fix-spec-file:
+        - "bash .packit.sh"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "aardvark-dns"
+# This version specification is reused by .packit.sh to generate rpm version
 version = "1.6.0-dev"
 edition = "2018"
 authors = ["github.com/containers"]


lsm5:
https://github.com/containers/netavark/pull/399#issuecomment-1233365737 applies here too

Luap99:
@lsm5 Is this still needed?

lsm5:
> @lsm5 Is this still needed?

@Luap99 yes, hoping to get this going soon. keep this open. Thanks!

lsm5:
@edsantiago @cevich wdyt? 

The copr build url is at: https://copr.fedorainfracloud.org/coprs/packit/containers-aardvark-dns-214/build/5433851/ . The package NVR lists the PR # as well as the head commit id in the release field.

We should be able to add centos stream as well if you're cool with proceeding.

lsm5:
This PR depends on https://github.com/containers/aardvark-dns/pull/288 .

lsm5:
Most packit builds seem to complete within the total time required by the cirrus tests, so hopefully that should not be a blocker.

Luap99:
> Most packit builds seem to complete within the total time required by the cirrus tests, so hopefully that should not be a blocker.

Assuming that this helps to catch bugs like https://github.com/containers/netavark/issues/578 early I think it is worth it even if it takes longer.

.packit.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - fedora-all-ppc64le
+      - fedora-all-s390x
+      - fedora-all-x86_64
+    srpm_build_deps:
+      - cargo
+      - make
+      - openssl-devel
+    actions:
+      post-upstream-clone:
+        - "curl -O https://src.fedoraproject.org/rpms/aardvark-dns/raw/rawhide/f/aardvark-dns.spec"
```
Luap99, line 22: Does this mean your downstream fedora spec file changes can break upstream CI?
I mean I agree that it is good so we can fix packaging bugs before we release but this is something we should be aware of.

Luap99 (commented):
LGTM

.packit.yaml
lsm5: I *think* we can use `aardvark-dns.spec.rpkg` from upstream instead, but gotta verify. That would mean no dependency on fedora dist-git.

lsm5: hmm, doesn't work yet with spec.rpkg files. I'll check with the packit people. Not depending on fedora dist-git would be super convenient.

lsm5:
hmm, dunno why custom copr project is taking so long.

.packit.yaml
```diff
@@ -0,0 +1,37 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
```
cevich, line 2: Thanks for referencing docs.

While this YAML may work, to be grammatically correct the file should begin with a `---` on the first line.  This is just a nit - no need to completely rebuild just for this small change.

.packit.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Packit's default fix-spec-file often doesn't fetch version string correctly.
+# This script handles any custom processing of the dist-git spec file and gets used by the
+# fix-spec-file action in .packit.yaml
+
+set -eo pipefail
+
+# Get Version from HEAD
+VERSION=$(grep '^version' Cargo.toml | cut -d\" -f2 | sed -e 's/-/~/')
```
cevich, line 10: At your next opportunity, I would suggest also adding a comment to the `Cargo.toml` file pointing back here.  This might avoid somebody inadvertently breaking this script by format changes in `Cargo.toml`.

cevich:
The script and YAML LGTM, with only a few nits.  I can't review the system as a whole because I'm packaging-dumb :grin:

.packit.yaml
lsm5: done

.packit.sh
lsm5: done

.packit.yaml
lsm5: @Luap99 now reuses the spec.rpkg file, so we're not dependent on official fedora package anymore.

lsm5:
disabled `fedora-36-armhfp` from the copr build targets as the builds are failing weirdly and I honestly don't think it's worth the effort as fedora 36 is lower priority and will soon go EOL. I've left a TODO on the copr homepage regardless.

lsm5:
@Luap99 @cevich @edsantiago PTAL again. Good to go, I hope.

lsm5:
@cevich if you're happy with this, maybe we can remove the `centos9_build` cirrus task. wdyt?

edsantiago:
OIC, the Makefile/CARGO commit was #290, which merged, which is why the commit disappeared. That took me way too long to figure out.

LGTM, with the caveat that I don't grok Packit. My quibble are minor:
- "Fix X" comments aren't really helpful, they'd be more useful if they said something like "Version derived from Cargo", "PACKIT_RPMSPEC_RELEASE" is given to us by packit, see http/foo", "Source0 is the tarball we generated above", etc. Basically, comments should say something that isn't obvious from the code itself. Not worth repushing for though.

lsm5:
> * "Fix X" comments aren't really helpful, they'd be more useful if they said something like "Version derived from Cargo", "PACKIT_RPMSPEC_RELEASE" is given to us by packit, see http/foo", "Source0 is the tarball we generated above", etc. Basically, comments should say something that isn't obvious from the code itself. Not worth repushing for though.

I'm happy to push again.

edsantiago:
Nah - the build is passing. Not worth it unless @cevich finds other concerns.

edsantiago (approved):
/lgtm
/hold

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "aardvark-dns"
+# This version specification is reused by .packit.sh to generate rpm version
```
cevich, line 3: Note: Renovate will muck with these values.  Assuming packit runs in all PRs, that will include the renovate proposed-updates as well.  If this is a concern, it's possible to teach renovate to ignore certain dependencies and never try to update them.

cevich (approved):
LGTM

Cargo.toml
lsm5: do you have an example in mind?

cevich: Renovate manages dependency versions.  It periodically checks dependencies using language-idiomatic methods, i.e. cargo.conf for rust.  When it finds an update, it modifies the file to bump the version number, and opens a PR with the change.

I mentioned it here because Renovate will most certainly not read or respect your comment :wink: So if modifying a version in this file could have unintended packit-consequences, I need to deal with that in the renovate configuration.  Is this okay, or do we need to inform renovate to keep it's hands off some dependency or another?

Luap99:
Lets get this in.
/hold cancel

Cargo.toml
lsm5: ah I see, let me send another PR to fix up the comments, there are some in .packit.sh too.

PR #220, opened 2022-09-06
run,serve: create aardvark `pid` in child before we notify parent process

flouthoc:
aardvark-dns must complete writing `pid` in config directory before
parent can return, hence notify parent only after child is done writing
the `pid` file.


Luap99 (approved):
LGTM

PR #229, opened 2022-09-27
Make build date reproducible

jelly:
To make aardvark-dns bit by bit reproducible the embedded build date needs to adhere to the SOURCE_DATE_EPOCH standard. So a rebuilder can set SOURCE_DATE_EPOCH to the build binary's build date and reproduce the binary succesfully.

https://reproducible-builds.org/docs/source-date-epoch/

Signed-off-by: Jelle van der Waa <jelle@archlinux.org>

---

Our reproducible test output before this change: https://reproducible.archlinux.org/api/v0/builds/343291/diffoscope

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,5 @@
-use chrono::Utc;
+use chrono::{DateTime, NaiveDateTime, Utc};
+use std::env;
 use std::process::Command;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


flouthoc:
@jelly Could you please run `cargo fmt`

jelly:
> @jelly Could you please run `cargo fmt`

Woops, sorry!

build.rs
```diff
@@ -1,12 +1,20 @@
...
 
 fn main() {
     // Generate the default 'cargo:' instruction output
     println!("cargo:rerun-if-changed=build.rs");
 
     // get timestamp
-    let now = Utc::now();
+    let now = match env::var("SOURCE_DATE_EPOCH") {
+        Ok(val) => {
+            let naive = NaiveDateTime::from_timestamp(val.parse::<i64>().unwrap(), 0);
```
flouthoc, line 12: Nit: The unwrap here will cause build to exit with error so maybe error can be handled here, but since this is only used while performing build so this is looks fine and not worth a repush.

flouthoc (approved):
Thanks @jelly , PR LGTM just one non-blocking ( not worth a repush ) nit.

build.rs
Luap99: I think unwrap is fine in the build file, I mean what do we want to do with the error anyway.

Luap99:
@jelly Can you open the same patch against netavark please?

jelly:
> @jelly Can you open the same patch against netavark please?

Sure!

PR #234, opened 2022-09-28
Cirrus: CI branch updates

cevich:
This Depends on https://github.com/containers/netavark/pull/423 merging and [a branch-build running and passing](https://cirrus-ci.com/github/containers/netavark/v1.1.0-rhel).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,6 @@ mock-rpm:
 .PHONY: validate
 validate: $(CARGO_TARGET_DIR)
 	cargo fmt --all -- --check
-	cargo clippy -p aardvark-dns -- -D warnings
 
 .PHONY: vendor
 vendor: ## vendor everything into vendor/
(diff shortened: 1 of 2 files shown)


lsm5:
> This Depends on #234 merging and [a branch-build running and passing](https://cirrus-ci.com/github/containers/netavark/v1.1.0-rhel).

this is self-referencing unless you meant a different PR

Makefile
```diff
@@ -98,7 +98,6 @@ mock-rpm:
 .PHONY: validate
 validate: $(CARGO_TARGET_DIR)
 	cargo fmt --all -- --check
-	cargo clippy -p aardvark-dns -- -D warnings
```
lsm5, line 101: does cargo fmt handle clippy too ?

Luap99:
LGTM but we should really pin the rust version to prevent the lint problems, especially on the maintenance branches.

cevich:
> this is self-referencing unless you meant a different PR

Woops...okay I fixed it.

> LGTM but we should really pin the rust version to prevent the lint problems, especially on the maintenance branches.

Yes, agreed, there is [an open issue for this](https://github.com/containers/automation_images/issues/150) so it shouldn't be forgotten.

Luap99 (approved):
/lgtm

PR #240, opened 2022-10-17
aardvark-dns: add support for container's custom `dns_servers`

flouthoc:
Aardvark-dns's config now allows containers to define a list of one of more custom `DNS` servers. If custom `DNS` servers are specified for a container aardvark-dns will use these custom DNS servers are resolver instead of host's default revolvers.
* Newly added field to aardvark-dns's config is optional in nature.

### How to verify
 - See newly added `integration` and `unit` tests.
 - Makes `podman run -it --rm --network test --dns 8.8.8.8 ctr dig google.com` to use `8.8.8.8` as a resolver instead of host's configured resolver.

### DIFF
--- a/src/test/config/podman_custom_dns_servers/podman
+++ b/src/test/config/podman_custom_dns_servers/podman
@@ -1,0 +1,5 @@
+10.88.0.1
+68fb291b0318b54a71f6f3636e58bd0896f084e5ba4fa311ecf36e019c5e6e43 10.88.0.2  condescendingnash 8.8.8.8
+68fb291b0318b54a71f6f3636e58bd0896f084e5ba4fa311ecf36e019c5e6e48 10.88.0.5  HelloWorld 3.3.3.3,1.1.1.1,::1
+95655fb6832ba134efa66e9c80862a6c9b04f3cc6abf8adfdda8c38112c2c6fa 10.88.0.3  hopefulmontalcini,testdbctr
+8bcc5fe0cb09bee5dfb71d61503a87688cfc82aa5f130bcedb19357a17765926 10.88.0.4  trustingzhukovsky,ctr1,ctra
--- a/src/test/test.rs
+++ b/src/test/test.rs
@@ -21,6 +21,11 @@ mod tests {
         }
     }
     #[test]
+    // Test loading of config file from directory with custom DNS for containers
+    fn test_loading_config_file_with_dns_servers() {
+        config::parse_configs("src/test/config/podman_custom_dns_servers").unwrap();
+    }
+    #[test]
     // Parse config files from stub data
     fn test_parsing_config_files() {
         match config::parse_configs("src/test/config/podman") {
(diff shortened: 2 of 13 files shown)


flouthoc:
Looks like I'm facing chicken-egg problem with integration tests, `netavark` needs new `aardvark` and `aardvark` needs new `netavark`.

mheon:
We've done this at least once before. How did we handle it then?

flouthoc:
> We've done this at least once before. How did we handle it then?

@mheon That was only aardvark-change and netavark was blocked on it but no issue at all here as well, this is sorted here https://github.com/containers/netavark/pull/452#discussion_r996967006 , we will get netavark PR in first then we can merge this and after that i'll unskip the netavark test in a followup PR.

flouthoc:
@Luap99 @mheon @baude PTAL this is ready for review and newly added integration tests pass with `netavark` from main.

test/helpers.bash
```diff
@@ -355,7 +355,8 @@ function run_in_host_netns() {
 # second arg is container_id
 # third is container name
 # fourth is subnet
-# fifth and greater are aliases
+# fifth is custom dns server for container (empty will not be used)
+# sixth and greater are aliases
 function create_config() {
```
Luap99, line 360: Nit, this is IMO not really maintainable. Adding arguments like will make it very hard to review and easy for callers to use it wrong. I would prefer an approach like this: https://github.com/containers/netavark/blob/d9e76ab68726cef40782d99f8e9df6e1d6d5c41f/test/helpers.bash#L325-L333
@edsantiago Do you have other suggestion to make this better?

flouthoc: I'm fine to convert this to `named` arguments and parse it or something even better that @edsantiago suggests.

src/test/config/podman_custom_dns_servers/podman
```diff
@@ -0,0 +1,5 @@
+10.88.0.1
+68fb291b0318b54a71f6f3636e58bd0896f084e5ba4fa311ecf36e019c5e6e43 10.88.0.2  condescendingnash 8.8.8.8
```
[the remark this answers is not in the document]
flouthoc: added in below line i.e `::1`

test/helpers.bash
edsantiago: @Luap99's suggestion looks great, and includes suitable error checking. I can't think of a better approach. (And, yes, I agree that positional args break down after about five).

flouthoc: SGTM. I'll amend this test.

flouthoc:
This is ready for review @Luap99 @baude @mheon PTAL

test/100-basic-name-resolution.bats
```diff
@@ -5,11 +5,61 @@
...
-@test "basic container - dns itself" {
+# custom DNS server is set to `3.3.3.3` which is invalid DNS server
+# hence all the external request must fail, this test is expected to fail
+# with exit code 124
+@test "basic container - dns itself (custom bad dns server)" {
 	setup_slirp4netns
 
 	subnet_a=$(random_subnet 5)
-	create_config "podman1" $(random_string 64) "aone" "$subnet_a" "a1" "1a"
+	create_config network_name="podman1" container_id=$(random_string 64) container_name="aone" subnet="$subnet_a" custom_dns_server="10.10.10.10" aliases='"a1", "1a"'
```
Luap99, line 15: your comment says 3.3.3.3 but here you have 10.10.10.10, both of them could become valid dns servers in the future,
I recommend to use 10.0.2.0 which is in the slirp4netns subnet and thus should not be routed elsewhere.

test/100-basic-name-resolution.bats
```diff
@@ -5,11 +5,61 @@
...
+	assert "$ip_a1"
+	# Set recursion bit is already set if requested so output must not
+	# contain unexpected warning.
+	assert "$output" !~ "WARNING: recursion requested but not available"
+
+        # custom dns server is set to 3.3.3.3 which is not a valid DNS server so external DNS request must fail
+	expected_rc=124 run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw"
+}
+
+# custom DNS server is set to `8.8.8.8` which is invalid DNS server
```
Luap99, line 31: should say valid?

test/100-basic-name-resolution.bats
```diff
@@ -5,11 +5,61 @@
...
+# hence all the external request must paas.
+@test "basic container - dns itself (custom good dns server)" {
+	setup_slirp4netns
+
+	subnet_a=$(random_subnet 5)
+	create_config network_name="podman1" container_id=$(random_string 64) container_name="aone" subnet="$subnet_a" custom_dns_server="8.8.8.8" aliases='"a1", "1a"'
```
Luap99, line 37: please add another test with two upstream servers

src/test/test.rs
```diff
@@ -21,6 +21,14 @@ mod tests {
         }
     }
     #[test]
+    // Test loading of config file from directory with custom DNS for containers
+    fn test_loading_config_file_with_dns_servers() {
+        match config::parse_configs("src/test/config/podman_custom_dns_servers") {
+            Ok(_) => {}
+            Err(e) => panic!("{}", e),
+        }
```
Luap99, line 29: why not just call unwrap()?

Luap99 (commented):
code LGTM, just some comments for the tests

test/100-basic-name-resolution.bats
flouthoc: fixed

test/100-basic-name-resolution.bats
flouthoc: fixed

test/100-basic-name-resolution.bats
flouthoc: fixed

src/test/test.rs
flouthoc: fixed for this test, will open different PR for older tests.

Luap99 (commented):
LGTM

Issue #249, opened 2022-11-04
dns: inbuilt resolver should return both `IPv6` and `IPv4` records if request type is `ANY`


flouthoc:
Inbuilt resolver returns `IPv4` and `IPv6` correctly for `A` or `AAAA` requests but it should return both when request type is `ANY`.

`ANY` request is not common but certain libraries and tools use this to get both `v6` and `v4` records therefore `aardvark-dns` should support it.

flouthoc:
Created a PR for this: https://github.com/containers/aardvark-dns/pull/250

tripougnif:
To support the default behaviour of most resolvers, aardvark-dns must also be able to respond to ANY queries over TCP.

PR #250, opened 2022-11-04
coredns: request must return both `IPv6` and `IPv4` for recordtype `ANY`

flouthoc:
Aardvark-dns must return both `IPv6` and `IPv4` if the DNS request has record type `ANY` instead of `A` or `AAAA` record type.

Closes: https://github.com/containers/aardvark-dns/issues/249


Luap99:
@flouthoc What is the state here? Can you add a test?

Luap99:
@flouthoc ping, is still still wanted?

flouthoc:
@Luap99 Not wanted by anybody but I just found it while playing, let me see if i can still repro this and the patch fixes the issue or not, I think there were some edgecase which i was hitting last time playing with this.

tripougnif:
> @Luap99 Not wanted by anybody but I just found it while playing, let me see if i can still repro this and the patch fixes the issue or not, I think there were some edgecase which i was hitting last time playing with this.

Nextcloud is doing ANY requests. I don't know if this is something they should fix on their end, but since other resolvers return A and AAAA records on ANY requests, I'd say this fix is needed.

tripougnif:
There are actually to issues. The first one is resolved with this PR. The second aardvark-dns is not bound to TCP and that ANY queries are going through TCP by default.

Luap99:
@flouthoc Still interested in finishing this? 
In the meantime tcp support was added and I reworked parts of this code heavily but I guess adding this should still be simple.

PR #251, opened 2022-11-04
 use cargo-vendor-filterer to create vendor tarball

Luap99:
This tool creates a much small vendor dir since we can only include the linux dependencies. Our current one on main is over 70 MB big and with cargo vendor-filterer it is only 6 MB.

see https://github.com/coreos/cargo-vendor-filterer


Luap99:
@cgwalters @baude @mheon @flouthoc PTAL, same as https://github.com/containers/netavark/pull/456

flouthoc (approved):
LGTM

PR #252, opened 2022-11-10
config,coredns: add support for `network scoped dns servers`

flouthoc:
Aardvark-dns users must be able to specify `dns_servers` for containers at network level as well, so all the containers must inherit custom dns servers from their network if specified.

What happens if both container's dns server and network's dns server are specified ?

- In such case priority order is `network dns servers` and then container's `custom dns servers`.

By default if nothing is specified resolution will happen from host's resolver


flouthoc:
Don't think failing test is related to this PR, I have restarted the test just to be sure.

flouthoc:
Don't review this yet intentionally breaking tests to see how config looks in failing `aarch64` integration test.

flouthoc:
Okay nevermind it was CI glitch this is fixed now and ready for review. We can either merge this or wait for me to implement e2e feature across netavark and podman. @Luap99 @mheon @baude PTAL

Luap99:
Did we already agreed on doing this now? I don't find anything in my mail or jira.

mheon:
Wait, why is priority order network servers and then container servers? that doesn't make any sense.

Container-specified servers should always have priority.

baude:
> Did we already agreed on doing this now? I don't find anything in my mail or jira.

We did not ... no decision has been rendered.

flouthoc:
Priority order can be reversed its not strict but lets hold this and make consensus. /hold

flouthoc:
PTAL this is needed for https://github.com/containers/netavark/pull/497

src/config/mod.rs
```diff
@@ -281,20 +308,32 @@ fn parse_config(path: &std::path::Path) -> Result<(Vec<IpAddr>, Vec<CtrEntry>),
...
+                            }
+                        };
+                    dns_server.extend(&ctr_dns_server);
                 }
+                // If network dns servers were specified then all the
+                // containers in this specified network must inherit
+                // network dns servers.
+                if !network_dns_servers.is_empty() {
+                    dns_server.extend(&network_dns_servers);
+                }
```
Luap99, line 332: Doesn't this just duplicate all network dns addresses for each container entry? This looks extremely inefficient. And I don't see how this can work with the next network update changes.

src/test/test.rs
```diff
@@ -82,6 +88,52 @@ mod tests {
...
+
+                // Shoudld not contain any DNS server other than network's dns server
+                dns_server = backend
+                    .ctr_dns_server
+                    .get(&IpAddr::V4(Ipv4Addr::new(10, 88, 0, 3)));
+                expected_dns_server = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+                assert_eq!(dns_server.unwrap().clone().unwrap()[0], expected_dns_server);
+                expected_dns_server = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 2));
+                assert_eq!(dns_server.unwrap().clone().unwrap()[1], expected_dns_server);
+            }
```
Luap99, line 132: Please rewrite this an a way that is more readable. Some for loop instead of the big amount of duplication would help a lot.

src/config/mod.rs
flouthoc: @Luap99 Yes we can optimize this further if add a new mapping in backend i'd like to do that later on but can do it now as well if that's a concern.

> And I don't see how this can work with the next network update changes.

Entire backend is shared with all the threads so its recreated from the config on this disk on every SIGHUP. So on network update `netavark` will just rewrite the `config` file and send `SIGHUP` to aardvark it will do the needful. But I'll verify refresh model once again manually.

src/test/test.rs
flouthoc: Adding loop SGTM

src/config/mod.rs
flouthoc: > Yes we can optimize this further.

Will do it in the next push.

flouthoc: Done.

flouthoc:
@Luap99 PTAL again.

src/config/mod.rs
```diff
@@ -43,6 +43,7 @@ pub fn parse_configs(
     let mut listen_ips_4: HashMap<String, Vec<Ipv4Addr>> = HashMap::new();
     let mut listen_ips_6: HashMap<String, Vec<Ipv6Addr>> = HashMap::new();
     let mut ctr_dns_server: HashMap<IpAddr, Option<Vec<IpAddr>>> = HashMap::new();
+    let mut network_dns_server: HashMap<String, Option<Vec<IpAddr>>> = HashMap::new();
```
Luap99, line 46: seems kinda point less to have an extra level of Option<>, I would just use `HashMap<String, Vec<IpAddr>>`.
You are already only adding the key if you have servers so the None value is not used.

Luap99 (approved):
LGTM

src/dns/coredns.rs
```diff
@@ -133,15 +133,26 @@ impl CoreDns {
...
                             if let Some(Some(dns_servers)) = self.backend.ctr_dns_server.get(&src_address.ip()) {
                                     if !dns_servers.is_empty() {
-                                        let mut nameservers_scoped: Vec<ScopedIp> = Vec::new();
                                         for dns_server in dns_servers.iter() {
                                             nameservers_scoped.push(ScopedIp::from(*dns_server));
                                         }
+                                    }
+                            }
+                            // Add network scoped resolvers
+                            if let Some(network_dns_servers) = self.backend.get_network_scoped_resolvers(&src_address.ip()) {
```
mheon, line 146: This should not be done if there were container-configured resolvers. We want container-provided resolvers, or network-provided resolvers. Not both.

config.md
```diff
@@ -4,9 +4,17 @@ Aardvark-dns will read configuration files from a given directory.
 
 Inside this directory there should be at least one config file. The name of the file equals the network name.
 
+### First line
 The first line in the config must contain a comma separated list of listening ips for this network, usually the bridge ips.
 At least one ip must be given.
+**Note**: Optional second coloumn allows users to specifify DNS servers at network level (seperated by comman), all the containers
```
baude, line 10: An optional second column of comma delimited domain name servers can be used at the network level.  All containers on that network will inherit all the specified name servers instead of using the host's resolver.

config.md
```diff
@@ -4,9 +4,17 @@ Aardvark-dns will read configuration files from a given directory.
...
+will inherit following nameservers instead of using host's resolver.
```
baude, line 12: you show a nice example below of what this looks like in the container.  is an example applicable here too?

src/backend/mod.rs
```diff
@@ -95,6 +99,27 @@ impl DNSBackend {
...
+    // Returns list of network resolvers for a particular container
+    pub fn get_network_scoped_resolvers(&self, requester: &IpAddr) -> Option<Vec<IpAddr>> {
+        let mut results: Vec<IpAddr> = Vec::new();
+
+        let nets = match self.ip_mappings.get(requester) {
+            Some(n) => n,
+            None => return None,
+        };
+
+        for net in nets {
```
baude, line 111: does it make sense to combine these two stanzas of code?

src/config/mod.rs
```diff
@@ -218,6 +252,24 @@ fn parse_config(path: &std::path::Path) -> Result<(Vec<IpAddr>, Vec<CtrEntry>),
                 bind_addrs.push(local_ip);
             }
 
+            // If network parts contain more than one col then
+            // we have custom dns server also defined at network level
+            // lets process that.
+            if network_parts.len() > 1 {
```
baude, line 258: does it make sense to invert this logic, and then put lines 221-222 in the conditional?  And then move the iteration outside the conditional?

baude:
fix the docs one for sure, the rest can be optional ...

config.md
flouthoc: Done.

config.md
flouthoc: I included all the examples in the last just like we do it in podman , adding another example here look redundant in docs. But I can add it if its needed.

src/dns/coredns.rs
```diff
@@ -133,15 +133,28 @@ impl CoreDns {
...
                             if let Some(Some(dns_servers)) = self.backend.ctr_dns_server.get(&src_address.ip()) {
                                     if !dns_servers.is_empty() {
-                                        let mut nameservers_scoped: Vec<ScopedIp> = Vec::new();
                                         for dns_server in dns_servers.iter() {
                                             nameservers_scoped.push(ScopedIp::from(*dns_server));
                                         }
+                                    }
+                            }
+                            // Add network scoped resolvers only if container specific resolvers were not configured
+                            if nameservers_scoped.is_empty() {
```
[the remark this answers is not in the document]
flouthoc: Thanks done.

mheon:
LGTM

flouthoc:
@baude PTAL

PR #263, opened 2022-12-06
release,make: add rule `crate-publish` for publishing crates.io

flouthoc:
* Add new rule `make crate-publish` to automate publish at https://crates.io/crates/aardvark-dns
* Document steps for publishing to `crates.io`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ LIBEXECDIR ?= ${PREFIX}/libexec
 LIBEXECPODMAN ?= ${LIBEXECDIR}/podman
 
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
+# Get crate version by parsing the line that starts with version.
+CRATE_VERSION ?= $(shell grep ^version Cargo.toml | awk '{print $$3}')
+GIT_TAG ?= $(shell git describe --tags)
 
 # Set this to any non-empty string to enable unoptimized
 # build w/ debugging features.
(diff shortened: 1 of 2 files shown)


Luap99:
/hold
see concerns in https://github.com/containers/netavark/pull/514#pullrequestreview-1206662087, lets keep the discussion there.

flouthoc:
@cevich Could it be that cirrus is down ? I see weird `404` from cirrus in CI logs.

cevich:
> @cevich Could it be that cirrus is down ? I see weird 404 from cirrus in CI logs.

``curl: (22) The requested URL returned error: 404``

Unlikely it's down.  More likely we're asking for the wrong URL or every job on the branch has failed.  Checking it out...

flouthoc:
> ...Oh interesting. This is my fault in a way. The permalink we use in CI here always tries to download from "The last successful job on main". What looks like happened is, [the latest job on main was successful](https://cirrus-ci.com/build/5461258955128832) but [did not post any build artifacts](https://cirrus-ci.com/task/5507773450944512) (because it was a `[skip-ci]` job). This seems like a Cirrus-CI bug, I will report it to them but I'm not sure if they can fix it or not.
> 
> As a workaround, you can merge a small docs update PR into netavark, so a job fully runs on main and publishes artifacts. That will resolve the problem probably faster than Cirrus-support can fix it. If they can't fix it, then we're screwed. I don't think there's anywhere I can't add a CI job to check for the presence of a `[skip-ci]` title wink We'll have to remember (ugg) to never ever use that magic string.

@cevich lol netavark is waiting for aardvark url https://github.com/containers/netavark/pull/514

cevich:
Issue in Cirrus-CI should now be fixed.

flouthoc:
Thanks @cevich , @Luap99 @baude @mheon PTAL

PR #267, opened 2022-12-07
Kill the chicken and egg part 2

baude:
Reliable CI cannot depend on another task always working when there is a chicken and egg dependency between them., Just build the netavark binary.

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -19,19 +19,11 @@ msg "Setting up runtime environment"
 msg "************************************************************"
 show_env_vars
 
-req_env_vars NETAVARK_URL
 
 set -x  # show what's happening
-curl --fail --location -o /tmp/netavark.zip "$NETAVARK_URL"
 mkdir -p /usr/libexec/podman
-cd /usr/libexec/podman
-unzip -o /tmp/netavark.zip
-if [[ $(uname -m) != "x86_64" ]]; then
-    mv netavark.$(uname -m)-unknown-linux-gnu netavark
-fi
-chmod a+x /usr/libexec/podman/netavark
-# show netavark commit in CI logs
-/usr/libexec/podman/netavark version
+cargo install --root /usr/libexec/podman --git https://github.com/containers/netavark
+mv /usr/libexec/podman/bin/netavark /usr/libexec/podman/
 
 # Warning, this isn't the end.  An exit-handler is installed to finalize
 # setup of env. vars.  This is required for runner.sh to operate properly.


cevich (dismissed):
The root cause is: using [skip-ci] on an aardvark or netavark PR. I've already contacted Cirrus-support about this two hours ago re: skip-ci merges should not be reflected in artifact download permalinks.

The issue can easily be worked around simply by merging some docs-change. All that's needed is a build on main that produces artifacts. Merging this PR will have unintended consequences for all netavark and aardvark CI.

cevich (commented):
Similar to https://github.com/containers/netavark/pull/518#pullrequestreview-1208814762

contrib/cirrus/setup.sh
```diff
@@ -19,19 +19,11 @@ msg "Setting up runtime environment"
...
-cd /usr/libexec/podman
-unzip -o /tmp/netavark.zip
-if [[ $(uname -m) != "x86_64" ]]; then
-    mv netavark.$(uname -m)-unknown-linux-gnu netavark
-fi
-chmod a+x /usr/libexec/podman/netavark
-# show netavark commit in CI logs
-/usr/libexec/podman/netavark version
+cargo install --root /usr/libexec/podman --git https://github.com/containers/netavark
+mv /usr/libexec/podman/bin/netavark /usr/libexec/podman/
```
cevich, line 26:
```suggestion
req_env_vars NETAVARK_URL

if showrun curl --fail --location -o /tmp/netavark.zip "$NETAVARK_URL"; then
    cd /usr/libexec/podman
    unzip -o /tmp/netavark.zip
    if [[ $(uname -m) != "x86_64" ]]; then
        mv netavark.$(uname -m)-unknown-linux-gnu netavark
    fi
    showrun chmod a+x /usr/libexec/podman/netavark
    # show netavark commit in CI logs
    /usr/libexec/podman/netavark version
else
    cargo install --root /usr/libexec/podman --git https://github.com/containers/netavark
    mv /usr/libexec/podman/bin/netavark /usr/libexec/podman
fi

```

cevich (commented):
Similar to https://github.com/containers/netavark/pull/518#pullrequestreview-1208814762

contrib/cirrus/setup.sh
cevich: (confirmed the cargo-install does work for arm64)

cevich:
My intention behind the `if...else...fi` suggestion is that we avoid doing a rust compile at test-time, wasting valuable developer brain-cells.  However, another potentially simpler solution just occurred to me:  Couldn't we just install netavark/aardvark RPMs from @lsm5 podman-next COPR?

cevich:
nevermind...won't work on (new) release-branches.

cevich:
-sigh- let me spare poor @baude brain-cells and just work on my own PRs for this, using the cargo-fallback idea...

PR #275, opened 2023-01-02
cargo: bump trust-dns-client and trust-dns-proto to 0.22

siretart:


flouthoc (commented):
Something seems off with PR all the tests fail for me on local with this

flouthoc:
@siretart Do you mind if i take over this PR ? I am experimenting with this and i see some issues.

siretart:
No not at all, please do drive this home!

flouthoc:
Okay I think this should pass now and should also fix the issues we are seeing on intermittent drops on some setups, I think it was issue way we were using sender stream and newer version of `trust-dns` have forced to creat dedicated stream for each peer.

Luap99:
> Okay I think this should pass now and should also fix the issues we are seeing on intermittent drops on some setups, I think it was issue way we were using sender stream and newer version of `trust-dns` have forced to creat dedicated stream for each peer.

Lets hope for the best. Code LGTM if it passes the tests.

Thanks @siretart and @flouthoc

flouthoc:
okay tests are green but hold let me becnhmark this as well.
/hold

flouthoc:
/hold cancel 

okay this is good to go . @mheon @baude PTAL

src/dns/coredns.rs
```diff
@@ -124,7 +127,8 @@ impl CoreDns {
                         Ok(msg) => {
                             let src_address = msg.addr();
                             let mut dns_resolver = self.resolv_conf.clone();
-                            let sender = sender.clone();
+                            let sender = sender_original.clone();
+                            let sender = sender.with_remote_addr(src_address);
```
mheon, line 131: Any reason this can't be combined into line 130?

flouthoc: done.

flouthoc:
@containers/netavark-maintainers PTAL

siretart:
I took the liberty of uploading this code change to Debian as [1.4.0-2](https://packages.debian.org/source/testing/aardvark-dns) because of https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1024980 -- and I have seen reports that this patch makes DNS resolution actually work for podman.

PR #312, opened 2023-03-31
use lower TTL for contianer entries

Luap99:
Containers can be short lived. We should not allow callers to cache the entry for to long. We could consider turning of cachine all together with a TTL of 0 but 60 seconds seems like a good tradeoff so the clients can still do some caching and not have to request the name every single time.

ref https://github.com/containers/netavark/discussions/644


baude:
yes LGTM

mheon:
Sure
/lgtm

PR #313, opened 2023-04-05
Releasev1.6.0

baude:
* dependency updates
* lower TTL for containers

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## v1.6.0
+* dependency updates
+* lower TTL for containers
+
 ## v1.5.0
 * dependency updates
 * code of conduct added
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## v1.6.0
+* dependency updates
+* lower TTL for containers
```
Luap99, line 5:
```suggestion
* lower the TTL to 60s for container names
```

maybe a bit more context would be helpful

Luap99 (commented):
one nit

RELEASE_NOTES.md
Luap99: you have to squash this commit into your release notes commit, right now it comes after the release commit so it will not be included in the tag.

Luap99:
Now my change is not applied at all?

PR #322, opened 2023-05-01
fix typo in makefile

baude:
the crate publish target in the makefile has a small typo where it was comparing two variables for version but one always had a prefix of `v`.


Luap99 (approved):
/lgtm

PR #343, opened 2023-06-16
run cargo update

Luap99:
Updates a lot of indirect dependencies, seems like renovate does not take care of it? I don't think dependabot did either so it's not a new thing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e579a7752471abc2a8268df8b20005e3eadd975f585398f17efcfd8d4927371"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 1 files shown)


Luap99:
@flouthoc @mheon PTAL

cc @cevich Any idea if we can get renovate to take care of this as well?

mheon:
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

cevich:
> Any idea if we can get renovate to take care of this as well?

It doesn't appear it can.  At the moment, the only "manager" module for Rust is called 'cargo'.  [The documentation for it](https://docs.renovatebot.com/modules/manager/cargo/#additional-information) suggests it reads `Cargo.toml`, then proposes updates w/ corresponding changes to the lock file.  The [Rust language documentation](https://docs.renovatebot.com/rust/) is even more clear.

Is it possible (and/or sensible) to make all the indirect deps into direct deps?  My guess is this isn't practicle.

Assuming there's a cargo command that can do this, another possible solution could be to use a github-actions workflow.  I think I came across a github action that can open PRs.  @Luap99 
 would you like me to investigate that more?

Luap99:
> Assuming there's a cargo command that can do this

As the title say `cargo update`, but there is no reason to over engineer this, I only run this once before we make a release every few months and it takes basically no time so I don't think it is a problem to do it manually. I was just asking because renovate has a lot of options so I thought there might be something.

cevich:
> I was just asking because renovate has a lot of options so I thought there might be something.

Nope, it was good of you to ask, and I'm happy to check.

PR #348, opened 2023-06-28
Releasev1.7.0

baude:


Cargo.lock
```diff
@@ -4,7 +4,7 @@ version = 3
 
 [[package]]
 name = "aardvark-dns"
-version = "1.7.0-dev"
+version = "1.7.0"
```
Luap99, line 7: this needs to be bumped to 1.8.0-dev as well

Luap99 (approved):
LGTM

PR #350, opened 2023-06-28
[CI:BUILD] RPM: cleanup spec and fix eln builds

lsm5:
- remove unnecessary conditionals
- fix ELN builds on copr environments

Note: The ELN build fix is actually handled on the COPR side by enabling the `local` repo. It's usually disabled by default on copr and enabled by default on koji, sometimes leading to inconsistent results.

### DIFF
--- a/rpm/aardvark-dns.spec
+++ b/rpm/aardvark-dns.spec
@@ -18,30 +18,8 @@
 %global debug_package %{nil}
 %endif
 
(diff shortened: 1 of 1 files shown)


lsm5:
good to go. PTAL

lsm5:
@Luap99 same thing as netavark PR. no worries if release commit merged already.

Issue #379, opened 2023-09-01
test_backend_network_scoped_custom_dns_server fails


vcunat:
100% reproducible for me, x86_64-linux (NixOS).
```
failures:

---- test::test::tests::test_backend_network_scoped_custom_dns_server stdout ----
thread 'test::test::tests::test_backend_network_scoped_custom_dns_server' panicked at 'assertion failed: `(left == right)`
  left: `["127.0.0.1", "::0.0.0.2"]`,
 right: `["127.0.0.1", "::2"]`', src/test/test.rs:110:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test::test::tests::test_backend_network_scoped_custom_dns_server

test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

Probably triggered by rustc: 1.71.1 -> 1.72.0, on a very quick glance.  I might bisect, in case you found that very useful.

flouthoc:
I'll check this thanks.

flouthoc:
PR: https://github.com/containers/aardvark-dns/pull/382 should close this.

PR #382, opened 2023-09-08
test: IPv6 format is changed in v1.72.0

flouthoc:
Output of IpAddr is changed in v`1.72.0` so tests must honor that.

Closes: https://github.com/containers/aardvark-dns/issues/379

### DIFF
--- a/src/test/test.rs
+++ b/src/test/test.rs
@@ -96,7 +96,10 @@ mod tests {
     fn test_backend_network_scoped_custom_dns_server() {
         match config::parse_configs("src/test/config/network_scoped_custom_dns") {
             Ok((backend, _, _)) => {
-                let expected_dnsservers = vec!["127.0.0.1", "::0.0.0.2"];
+                let expected_dnsservers = vec![
+                    IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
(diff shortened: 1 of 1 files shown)


src/test/test.rs
```diff
@@ -107,7 +106,13 @@ mod tests {
...
                         output_dnsservers.push(format!("{}", server));
                     }
-                    assert_eq!(expected_dnsservers, output_dnsservers);
+                    if rustc_version::version().unwrap()
+                        >= rustc_version::Version::parse("1.72.0").unwrap()
+                    {
+                        assert_eq!(vec!["127.0.0.1", "::2"], output_dnsservers);
+                    } else {
+                        assert_eq!(vec!["127.0.0.1", "::0.0.0.2"], output_dnsservers);
+                    }
```
Luap99, line 115: Wouldn't it be much better to just directly compare based on the actual IpAddr struct instead of the string representation, then we do not need to worry about this I think.

flouthoc: Sounds like a good idea, let me check that.

baude:
lgtm after you address the comment from @Luap99

Luap99 (approved):
/lgtm

PR #383, opened 2023-09-11
bump nix to 0.27.1

Luap99:
There are no longer any default features so we need to list explicitly what we use.


flouthoc (approved):
LGTM
/approve
/lgtm

PR #394, opened 2023-10-11
tmt: initial enablement

lsm5:


lsm5:
@cevich @edsantiago @baude @Luap99 @flouthoc @mheon PTAL

This PR enables all tests for x86_64 and aarch64 on fedora, centos stream and rhel nightlies. The integration tests use the binary built by the copr rpm task, so they take a little longer. We can set them to not depend on the rpm task if it takes annoyingly long.

Luap99:
Why do we need so many distros to run unit an integration tests on? These seems like a big waste of compute time.

Compile tests for RHEL seems like a good idea given https://github.com/containers/netavark/issues/775 but I don't see the point to run unit and integration tests on so many distros, once for x86_64 and aarch64 should be good enough IMO.

Also if we use tmt then we must configure the merge bot to actually block on these tests (it is not the case in podman so I am not sure if it is the case here). But I guess the merge bot is broken either way and allows merges without passing tests so that doesn't matter really to prevent early merges.

Luap99:
Even more important depending on dnf install at runtime will causes al lot of headaches down the road. Right now we static well known VM images that must be explicitly updated.
With this PR we get new compile/linting problems or general bugs in dependencies on random PRs that have nothing to do with it. And that is extremely annoying to deal with as developer.

cevich:
First off, thank you @lsm5 for taking this on and experimenting with it.  It's certainly something worthwhile, since I don't pay much attention outside of maintaining our current (possibly inferrior?) setup.

> These seems like a big waste of compute time.

Compute time is cheap, that doesn't bother me.  It's the potential for increased developer time/frustration I mostly have on my mind.

> but I don't see the point to run unit and integration tests on so many distros

I also agree from the perspective of limiting the potential for flakes.  Again, those are a huge time/frustration hole for developers.

> Also if we use tmt then we must configure the merge bot

The TMT setup has exactly the same problem as Cirrus that breaks the bot: Test names that change over time (they include release numbers).  Years ago when I investigated this in prow, there's no way to configure a wild-card match.  It's 100% exact name match or nothing and no in-between option.

There *IS* the possibility (and maybe need) to key the bot off of an aggregate "Success" task.  This is also handy because it clearly tracks which tests are REALLY required to pass, and lets you know immediately if one never "spawned" for some reason.

> Right now we static well known VM images that must be explicitly updated.

Yes, this was done deliberately (again, surprise!) to save developer time/frustration.  IIRC, Lokesh said these "dynamic" environments are cached in some way.  But even with cache, the concern still applies IMHO. These environments will have both an increased chance of flakes and other unexpected outcomes due to dependence on runtime installation.

In all my experience working with various CI systems, runtime-updates are the one thing I can almost always point to as adding an unnecessary PITA factor.  But I'm also not one to overly criticize without trying to be constructive:

@lsm5 is it possible to hard-code each package version/release, and then we could (somehow) teach Renovate how to manage updates to them (maybe grouped all in one PR)?  Renovate can [read new versions from a pretty large list of sources](https://docs.renovatebot.com/modules/datasource/).

It's not a perfect solution (there would be some desync. from one repo. CI to another).  But maybe it's workable?

lsm5:
> Why do we need so many distros to run unit an integration tests on? These seems like a big waste of compute time.

Not our money. The people that maintain it told me we don't have to worry about capacity for the time being. Regardless, we can easily trim the number of envs the tests are run on.

> 
> Compile tests for RHEL seems like a good idea given [containers/netavark#775](https://github.com/containers/netavark/issues/775) but I don't see the point to run unit and integration tests on so many distros, once for x86_64 and aarch64 should be good enough IMO.

IIUC, one of @edsantiago 's pain points was RHEL problems were found very late in the process. Either way, very easy to manage the test matrix and the integration tests don't wait on each other, so whatever test matrix we all can agree on is fine by me.

> 
> Also if we use tmt then we must configure the merge bot to actually block on these tests (it is not the case in podman so I am not sure if it is the case here). But I guess the merge bot is broken either way and allows merges without passing tests so that doesn't matter really to prevent early merges.

The tmt tests in podman so far are only reverse dependency tests on something outside our org and I didn't want to block on tmt until people had a good feel for it. I `think` just like cirrus we can create a `success` task and block on that once we're comfortable blocking on tmt and maybe removing cirrus.

> @lsm5 is it possible to hard-code each package version/release, and then we could (somehow) teach Renovate how to manage updates to them (maybe grouped all in one PR)? Renovate can [read new versions from a pretty large list of sources](https://docs.renovatebot.com/modules/datasource/).

@cevich see: https://tmt.readthedocs.io/en/stable/spec/plans.html#install . We can install packages using URLs, so hardcoding shouldn't be a problem. We may need to decide what exactly we're testing though, do aardvark's integration tests fetch netavark from the distro's official repos, or from podman-next to be as close to upstream main as possible? I guess we could do both if we need both.

Luap99:
> > Why do we need so many distros to run unit an integration tests on? These seems like a big waste of compute time.
> 
> Not our money. The people that maintain it told me we don't have to worry about capacity for the time being. Regardless, we can easily trim the number of envs the tests are run on.

Fair enough but I think they will care sooner or later and then we are in the same boat again. I see a lot of value in RHEL build tests as we depend on the rustc shipped there.

> > Compile tests for RHEL seems like a good idea given https://github.com/containers/netavark/issues/775 but I don't see the point to run unit and integration tests on so many distros, once for x86_64 and aarch64 should be good enough IMO.
>
> IIUC, one of @edsantiago 's pain points was RHEL problems were found very late in the process. Either way, very easy to manage the test matrix and the integration tests don't wait on each other, so whatever test matrix we all can agree on is fine by me.

I think that is great for podman where we have runtime dependencies (or even netavark where we make use a lot of kernel networking APIs) but aardvark-dns is a standalone server so there is very little benefit IMO.
Instead it will multiply the changes for flakes which will be annoying.



> The tmt tests in podman so far are only reverse dependency tests on something outside our org and I didn't want to block on tmt until people had a good feel for it. I think just like cirrus we can create a success task and block on that once we're comfortable blocking on tmt and maybe removing cirrus.

Yeah I think maybe I am missing the goal of this PR. I thought the point was to remove the current testing tasks.


---

Also for package versions netavark must be installed as latest version from main (the podman/next copr package should work fine), we cannot use the main distro versions as netavark and aardvark-dns must be kept in sync as we may change the API between them at any time.

Anyway I guess the biggest fear of mine would be the runtime breakages due whatever dnf update brings us. This would be a regular case because of the rustc updated which include new linter rules so the validate task would likely fail on each update.  I like the current way of having an explicit VM image update PR.

lsm5:
> I think that is great for podman where we have runtime dependencies (or even netavark where we make use a lot of kernel networking APIs) but aardvark-dns is a standalone server so there is very little benefit IMO. Instead it will multiply the changes for flakes which will be annoying.

Sure, fine by me. I'll be happy to update to a smaller test matrix. FWIW, we currently don't have integration tests running in either fedora bodhi updates or rhel gating for aardvark-dns and netavark.
 
> Yeah I think maybe I am missing the goal of this PR. I thought the point was to remove the current testing tasks.

Yes, that is one of the goals, though I don't know if we wanna do it in this PR itself or maybe let both TMT and cirrus live together for a while and then remove cirrus after some time. I'll defer to @edsantiago and @cevich .
 
> Also for package versions netavark must be installed as latest version from main (the podman/next copr package should work fine), we cannot use the main distro versions as netavark and aardvark-dns must be kept in sync as we may change the API between them at any time.

ack

> 
> Anyway I guess the biggest fear of mine would be the runtime breakages due whatever dnf update brings us. This would be a regular case because of the rustc updated which include new linter rules so the validate task would likely fail on each update. I like the current way of having an explicit VM image update PR.

we could `probably` address this with installing a fixed version-release of rustc via rpm, or via their curl way and not run any `dnf update` before the validate task. The `prepare` step can be moved around into the individual tests and doesn't have to be global.

Luap99:
If you like to merge this as experiment without blocking anything on it that would be fine for me. We can see how stable it is over the next weeks and then later remove the cirrus jobs and block on the tmt jobs once we are confident.

lsm5:
> If you like to merge this as experiment without blocking anything on it that would be fine for me. We can see how stable it is over the next weeks and then later remove the cirrus jobs and block on the tmt jobs once we are confident.

@Luap99 Thanks a lot. I'll wait to get acks from @cevich and @edsantiago

cevich:
> installing a fixed version-release of rustc via rpm

TMT should follow the setup currently used in c/automation_images.

Re: NV/AV sync:

TMT should be able to do this using the same cirrus-artifacts based mechanism, no?

Re: package updates and test-environment stability.

I found a [package-repo. datasource in renovate (supports Fedora + CentOS, but no RHEL)](https://docs.renovatebot.com/modules/datasource/repology/). When used with a regex-manager, it could handle managing package updates for the CI environment, substantially similar to what we do now with pre-build VM images.  It will be a brain-twister to configure it, but based on the example, seems possible.  It would need to be configured on a repo-by-repo basis, which could be a minor concern.

As an experiment, this all LGTM also.

cevich:
Followup: Can TMT scoop up artifact files and present them via permalink (i.e. "latest")?

Luap99:
>> Re: NV/AV sync:
>
> TMT should be able to do this using the same cirrus-artifacts based mechanism, no?

I think the already existent packit automation with the podman-next copr is enough, we do not need to copy files exactly.

cevich:
> we do not need to copy files exactly.

Ahh right, it's just using "good enough" RPMs.  k.

cevich:
@lsm5 As a follow-up, would you be willing to experiment with that _Repology_ datasource + regex manager (in this repo.) to see if it could replace our use of pre-built VM images in Cirrus while maintaining a developer-controlled update workflow?

I completely understand "no, Renovate is too hard".  Just thought I'd ask, since that solution could be generally applicable in an eventual Cirrus -> TMT migration IMHO.

lsm5:
> @lsm5 As a follow-up, would you be willing to experiment with that _Repology_ datasource + regex manager (in this repo.) to see if it could replace our use of pre-built VM images in Cirrus while maintaining a developer-controlled update workflow?

@cevich sure, mind creating a jira with the exact requirements you're looking for?

cevich:
Sure.

PR #402, opened 2023-10-18
aardvark main: change error reporting

Luap99:
Never report an error when the syslog init fails, there are systems without syslog (e.g. inside containers) and we should not spam a random expected messages in this context.

Also add aardvark-dns to the error context as this is printed to stderr which is inherited from the podman command so it is not clear which process wrote it otherwise.

Fixes https://github.com/containers/podman/issues/19809


edsantiago:
LGTM but I have no way to test it in podman CI. Thanks for addressing it.

PR #407, opened 2023-11-06
decouple internal testing farm tasks

lsm5:
This way non-internal testing farm tasks don't have to wait on approval from someone with commit access to the repo. Especially useful for update PRs created by Renovate.

The integration tests run on the internal testing farm are linked to this new build job using the same identifier `integration_test_internal`.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -33,6 +33,25 @@ jobs:
       - "copr://rhcontainerbot/podman-next"
     enable_net: true
 
+  # Jobs that run on internal testing farm are not run automatically for
+  # external contributors. Having a separate build task for consumption by
+  # internal testing farm tasks decouples the internal testing farm tasks from
+  # the non-internal ones so that non-internal tasks get run automatically.
+  - job: copr_build
+    trigger: pull_request
+    notifications:
+      failure_comment:
+        message: "Ephemeral COPR build failed. @containers/packit-build please check."
+    targets:
+      - epel-8-x86_64
+      - epel-8-aarch64
+      - epel-9-x86_64
+      - epel-9-aarch64
+    additional_repos:
+      - "copr://rhcontainerbot/podman-next"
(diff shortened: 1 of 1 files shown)


lsm5:
I guess this can only be verified after it's merged and renovate or a non-committer creates a PR. Also, this change should not create rpm-build:fedora-* tasks but right now they are getting created because of https://github.com/packit/packit-service/issues/2131 .

lsm5:
@Luap99 PTAL

Luap99:
There seems to be a bunch of duplicated tasks for the rpm-build tasks now??

lsm5:
> There seems to be a bunch of duplicated tasks for the rpm-build tasks now??

@Luap99 the ones for epel are expected, we need a separate build job for use by internal testing farm jobs. The duplicated tasks for fedora are being tracked in the packit issue.

Luap99:
But we do need twice the same job? IS there no way to make it only build once and have the RHEL task consume the right build?

Doing the exact same thing twice is totally wasting resource increasing odds for flakes and totally confusing when looking at this job list.

Luap99:
Or maybe asking the other way around, couldn't we drop the existing build taks then? I feel like I missing how this is supposed to work.

baude:
what say you @lsm5

lsm5:
This can be closed. With the merging of https://github.com/containers/aardvark-dns/pull/458 , the fedora jobs at least `should` run for all PRs regardless of the PR submitter being maintainer / outside contributor. I suspect the centos 9 jobs `should` run the same way as fedora jobs and only rhel9 jobs will need a maintainer to trigger them. Guess we'll find out in upcoming PRs. Worst case, a maintainer can trigger them all.

Luap99:
@lsm5 yes it seems to work if I look at https://github.com/containers/aardvark-dns/pull/456

PR #410, opened 2023-11-20
run cargo update

Luap99:


mheon:
Changes LGTM once you can get the tests happy

Luap99:
We do not block on tmt tests at all, these are an experiment for now.
That said the failure it reports is legit and can also fail in our normal tests because this is a flake that exist for quite some time now, I filed https://github.com/containers/aardvark-dns/issues/411 to track it.

flouthoc (approved):
LGTM
/lgtm
/approve

PR #413, opened 2023-11-30
server: remove unused kill switch

Luap99:
Ever since commit a49d70e we no longer use the kill switch, it is only logged in one trace statement but that does not add any value so just get rid of it. The broadcast channel handles this already much better.


flouthoc (approved):
LGTM

mheon:
The failure looks suspicious but I can't see how it's related to this PR. I'm assuming it's a flake and restarted it.

PR #427, opened 2024-01-29
[CI:DOCS] Packit: disable bodhi tasks

lsm5:
This allows manual submission of a single bodhi update containing netavark and aardvark-dns X.Y builds.

@Luap99 PTAL

Netavark will have a similar PR with additional dep handling in its spec.


Luap99 (approved):
/lgtm
/hold

PR #437, opened 2024-02-28
Add names for domains in answers in reverse lookups

### LINKED ISSUE #436: Reverse lookups in podman return `.` for domain name in answer section
When I run reverse lookups inside podman in docker compose with `aardvark-dns`, resolving `A` records works but resolving reverse lookups does not with hickory DNS.

I think the issue is because in the answer section the domain is just `.` instead of the queried domain but my knowledge of inner workings of DNS is limitted so I might be wrong here.

```
root@1e714d3fa668:/# dig -x 10.89.0.63

; <<>> DiG 9.18.24-1-Debian <<>> -x 10.89.0.63
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 48549
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 966cea63d68458dc (echoed)
;; QUESTION SECTION:
;63.0.89.10.in-addr.arpa.	IN	PTR

;; ANSWER SECTION:
.			60	IN	PTR	project-container-1.
.			60	IN	PTR	project-container-1.
.			60	IN	PTR	container.
.			60	IN	PTR	190053e6f903.

;; Query time: 0 msec
;; SERVER: 10.89.0.1#53(10.89.0.1) (UDP)
;; WHEN: Tue Feb 27 16:39:01 UTC 2024
;; MSG SIZE  rcvd: 160
```

When I run the same with docker, the answer section looks like this:
```
;; ANSWER SECTION:
8.0.20.172.in-addr.arpa. 600	IN	PTR	project-container-1.project-network.
```

mladedav:
Closes #436 
Hopefully...

I wasn't able to run `aardvark-dns` locally so I wasn't able to test it. Running `RUST_LOG=trace cargo run -- -c src/test/config/podman -p 5533 run` spawns the background process and the pid file is created but I am unable to query it. Any asistance with that would be appreciated.

I also fixed two typos I have noticed along the way.

### DIFF
--- a/test/500-reverse-lookups.bats
+++ b/test/500-reverse-lookups.bats
@@ -22,18 +22,22 @@ load helpers
 	create_container "$a2_config"
 	a2_pid="$CONTAINER_NS_PID"
 
-	echo "$a1_config"
-	echo "$a2_config"
+	echo "a1 config:\n${a1_config}\n"
+	echo "a2 config:\n${a2_config}\n"
 
 	# Resolve IPs to container names
 	dig_reverse "$a1_pid" "$a2_ip" "$gw"
-	assert "$output" =~ "atwo"
-	assert "$output" =~ "a2"
-	assert "$output" =~ "2a"
+	echo -e "Output:\n${output}\n"
+	a2_expected_name=$(echo $a2_ip | awk -F. '{printf "%d.%d.%d.%d.in-addr.arpa.", $4, $3, $2, $1}')
+	assert "$output" =~ "$a2_expected_name[ 	].*[ 	]atwo\."
+	assert "$output" =~ "$a2_expected_name[ 	].*[ 	]a2\."
+	assert "$output" =~ "$a2_expected_name[ 	].*[ 	]2a\."
 	dig_reverse "$a2_pid" "$a1_ip" "$gw"
-	assert "$output" =~ "aone"
-	assert "$output" =~ "a1"
-	assert "$output" =~ "1a"
+	echo -e "Output:\n${output}\n"
+	a1_expected_name=$(echo $a1_ip | awk -F. '{printf "%d.%d.%d.%d.in-addr.arpa.", $4, $3, $2, $1}')
+	assert "$output" =~ "$a1_expected_name[ 	].*[ 	]aone\."
+	assert "$output" =~ "$a1_expected_name[ 	].*[ 	]a1\."
+	assert "$output" =~ "$a1_expected_name[ 	].*[ 	]1a\."
 }
 
 @test "check reverse lookups on ipaddress v6" {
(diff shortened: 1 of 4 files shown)


mladedav:
/assign mheon

Sorry if I shouldn't have assigned you yet, the bot said I should do that after it has been reviewed, but it did not assign any reviewers so I'm not sure if it applies to this repo.

mladedav:
No, I have the latest `netavark-1.10.2-1` from arch repository, but I can see there is already `1.10.3` released. I'd hope the patch version would not be the issue?

I'll try full upgrade of the system and if it doesn't help I'll try to see if I can make it somehow work with virtualization or on another machine.

Luap99:
> No, I have netavark-1.10.2-1 from manjaro repository (latest packaged version there), but I can see there is already 1.10.3 released. I'd hope the patch version would not be the issue?

Yes that is more than new enough. I know we had switch were it would cause a hang but that was a while ago maybe 1.4 or 1.5 but 1.10.2 should work just fine.

Luap99:
Tested it locally, your fix seems fine to populate the ANSWER section correctly.
There is still the difference with the network name suffix but I don't think that should matter for your use case.

mladedav:
I've made the tests work locally. Turns out the issue was that my netavark is in a different place than the default variables expected and instead of failing the test just hangs in the `run timeout` command for some reason.

Anyway I've made the tests work and as far as I can see the answers look correct.

test/500-reverse-lookups.bats
```diff
@@ -22,18 +22,20 @@ load helpers
...
+	echo "a1 config:\n${a1_config}\n"
+	echo "a2 config:\n${a2_config}\n"
 
 	# Resolve IPs to container names
 	dig_reverse "$a1_pid" "$a2_ip" "$gw"
-	assert "$output" =~ "atwo"
-	assert "$output" =~ "a2"
-	assert "$output" =~ "2a"
+	echo -e "Output:\n${output}\n"
+	assert "$output" =~ '([0-9]+\.){4}in-addr\.arpa\.[ 	].*[ 	]atwo\.'
```
Luap99, line 31: I think we should make sure that the ip is printed in reverse order.
This regex would allow any ip but we really should make sure it is actually the ip we want.

Maybe for ipv4 we can do `echo $ip | awk -F. '{print $4"."$3"." $2"."$1}'` to invert it, for ipv6 something like this would not really work but I think if we at least properly verify ip4.
For ipv6 our test should always use a fd... adress so you could at least match .d.f.ipv6.arpa

test/helpers.bash
```diff
@@ -598,7 +598,7 @@ function dig_reverse() {
     # second arg is the IP address
     # third arg is server addr
     #run_in_container_netns "$1" "dig" "-x" "$2" "+short" "@$3"
```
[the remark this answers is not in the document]
mladedav: Done

test/500-reverse-lookups.bats
mladedav: Ok, I did what you suggested except I used `printf` and constructed the whole address. For IPv6 I also can't think of anything that would help much so I went with what you proposed.

Luap99 (approved):
sorry for the delay, LGTM
FYI we generally prefer code+test in the same commit but I won't block over that. 
@baude @mheon PTAL

PR #447, opened 2024-04-05
Internal networks cannot make external DNS requests

mheon:
Internal networks cannot connect to the internet thanks to routing, but they can connect to Aardvark, which will happily forward their DNS requests to the internet.

This could theoretically be used to build a data-exfiltration sidechannel.

Fix this by identifying internal networks with a filename suffix (using a character disallowed in actual network names to ensure we don't conflict with another network) and explicitly setting their DNS servers to an empty list (and refusing to set per-container DNS at all). We could actually error on finding DNS servers in an internal network, but silently ignoring prevents possible compatibility issues with Netavark.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -22,6 +22,8 @@ pub struct DNSBackend {
     pub ctr_dns_server: HashMap<IpAddr, Option<Vec<IpAddr>>>,
     // Map of network name and DNS server IPs.
     pub network_dns_server: HashMap<String, Vec<IpAddr>>,
+    // Map of network name to bool (network is/is not internal)
+    pub network_is_internal: HashMap<String, bool>,
 }
 
 pub enum DNSResult {
@@ -45,6 +47,7 @@ impl DNSBackend {
         reverse: HashMap<String, HashMap<IpAddr, Vec<String>>>,
         ctr_dns_server: HashMap<IpAddr, Option<Vec<IpAddr>>>,
         network_dns_server: HashMap<String, Vec<IpAddr>>,
+        network_is_internal: HashMap<String, bool>,
     ) -> DNSBackend {
         DNSBackend {
             ip_mappings: containers,
@@ -52,6 +55,7 @@ impl DNSBackend {
             reverse_mappings: reverse,
             ctr_dns_server,
             network_dns_server,
+            network_is_internal,
         }
     }
 
@@ -118,6 +122,31 @@ impl DNSBackend {
         };
 
         Some(results)
+    }
+
+    // Checks if a container is associated with only internal networks.
+    // Returns true if and only if a container is only present in
+    // internal networks.
+    pub fn ctr_is_internal(&self, requester: &IpAddr) -> bool {
+        match self.ip_mappings.get(requester) {
+            Some(nets) => {
+                for net in nets {
+                    match self.network_is_internal.get(net) {
+                        Some(internal) => {
+                            if !internal {
+                                return false;
+                            }
+                        }
+                        None => continue,
+                    }
+                }
+            }
+            // For safety, if we don't know about the IP, assume it's probably
+            // someone on the host asking; let them access DNS.
+            None => return false,
+        }
+
+        true
     }
 
     /// Return a single name resolved via mapping if it exists.
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -44,6 +44,7 @@ pub fn parse_configs(
     let mut listen_ips_6: HashMap<String, Vec<Ipv6Addr>> = HashMap::new();
     let mut ctr_dns_server: HashMap<IpAddr, Option<Vec<IpAddr>>> = HashMap::new();
     let mut network_dns_server: HashMap<String, Vec<IpAddr>> = HashMap::new();
+    let mut network_is_internal: HashMap<String, bool> = HashMap::new();
 
     // Enumerate all files in the directory, read them in one by one.
     // Steadily build a map of what container has what IPs and what
@@ -66,12 +67,20 @@ pub fn parse_configs(
                 }
                 let parsed_network_config = parse_config(cfg.path().as_path())?;
 
+                let mut internal = false;
+
                 let network_name: String = match cfg.path().file_name() {
                     // This isn't *completely* safe, but I do not foresee many
                     // cases where our network names include non-UTF8
                     // characters.
                     Some(s) => match s.to_str() {
-                        Some(st) => st.to_string(),
+                        Some(st) => {
+			    let name_full = st.to_string();
+			    if name_full.ends_with(constants::INTERNAL_SUFFIX) {
+				internal = true;
+			    }
+			    name_full.strip_suffix(constants::INTERNAL_SUFFIX).unwrap_or(&name_full).to_string()
+			},
                         None => return Err(std::io::Error::new(
                             std::io::ErrorKind::Other,
                             format!("configuration file {} name has non-UTF8 characters", s.to_string_lossy()),
@@ -85,11 +94,16 @@ pub fn parse_configs(
 
                 // Network DNS Servers were found while parsing config
                 // lets populate the backend
-                if !parsed_network_config.network_dnsservers.is_empty() {
+                // Only if network is not internal.
+                // If internal, explicitly insert empty list.
+                if !parsed_network_config.network_dnsservers.is_empty() && !internal {
                     network_dns_server.insert(
                         network_name.clone(),
                         parsed_network_config.network_dnsservers,
                     );
+                }
+                if internal {
+                    network_dns_server.insert(network_name.clone(), Vec::new());
                 }
 
                 for ip in parsed_network_config.network_bind_ip {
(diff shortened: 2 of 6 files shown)


src/config/mod.rs
```diff
@@ -124,7 +139,10 @@ pub fn parse_configs(
                                 .entry(IpAddr::V4(ip))
                                 .or_default()
                                 .append(&mut entry.aliases.clone());
-                            ctr_dns_server.insert(IpAddr::V4(ip), entry.dns_servers.clone());
+			    // DNS only accepted on non-internal networks.
+                            if !*network_is_internal.get(&network_name).unwrap_or(&bool::default()) {
+                                ctr_dns_server.insert(IpAddr::V4(ip), entry.dns_servers.clone());
+                            }
```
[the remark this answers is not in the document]
mheon: Ah, should be for v6 as well. I think that `podman run --net internal_net --dns 4.2.2.1 ...` or similar would still allow external DNS without this line.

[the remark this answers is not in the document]
Luap99: as yeah now that makes sense we have both the network and container level servers, now i understand what this map is use for.

test/100-basic-name-resolution.bats
```diff
@@ -165,3 +165,26 @@ load helpers
 	# contain unexpected warning.
 	assert "$output" !~ "WARNING: recursion requested but not available"
 }
+
+# Internal network, meaning no DNS servers.
+# Hence all external requests must fail.
+@test "basic container - internal network has no DNS" {
+	setup_slirp4netns
+
+	subnet_a=$(random_subnet 5)
```
Luap99, line 174: I See that all the other cases have this as well but it is wrong, for a ipv4 subnet do not call it with an argument, the only valid argument for this function is 6 for ipv6.

test/100-basic-name-resolution.bats
```diff
@@ -165,3 +165,26 @@ load helpers
 	# contain unexpected warning.
 	assert "$output" !~ "WARNING: recursion requested but not available"
 }
+
+# Internal network, meaning no DNS servers.
+# Hence all external requests must fail.
+@test "basic container - internal network has no DNS" {
```
Luap99, line 171: test seems logical, given the risk of regression we should also test the ipv6 path.

mheon:
Hm. I think we might need a patched Netavark in the VMs to make this work.

Luap99:
> Hm. I think we might need a patched Netavark in the VMs to make this work.

Well yeah that is the problem with the compatibility issues with nv/av.

Rethinking this I don't think the filename is the best idea given its compatibility issues on updates. What do you think about using a xattr on the file to indicate this? We would still need to update both, but if we just update one it will not cause any weird side effects because the old aardavrk-dns will just ignore it.

Luap99:
> Rethinking this I don't think the filename is the best idea given its compatibility issues on updates. What do you think about using a xattr on the file to indicate this? We would still need to update both, but if we just update one it will not cause any weird side effects because the old aardavrk-dns will just ignore it.

Nevermind, looks like tmpfs and "user." xattrs are rather new, it was just added last year so this would get us in trouble on older kernels.

src/config/mod.rs
```diff
@@ -66,12 +66,20 @@ pub fn parse_configs(
...
                     // cases where our network names include non-UTF8
                     // characters.
                     Some(s) => match s.to_str() {
-                        Some(st) => st.to_string(),
+                        Some(st) => {
+			    let name_full = st.to_string();
+			    if name_full.ends_with(constants::INTERNAL_SUFFIX) {
+				internal = true;
+			    }
+			    name_full.strip_suffix(constants::INTERNAL_SUFFIX).unwrap_or(&name_full).to_string()
```
flouthoc, line 81: Shouldn't `strip_suffix` be called on above bracket ( with `ends_with` ) only if a suffix was found ? Please ignore if this doesn't make sense.

flouthoc: Oh i only meant that it performs the search again, but yeah this is looks safe and I hope compiler will automatically remove redundant search. No no issues I guess.

src/config/mod.rs
```diff
@@ -66,12 +66,20 @@ pub fn parse_configs(
...
+                let mut internal = false;
+
                 let network_name: String = match cfg.path().file_name() {
                     // This isn't *completely* safe, but I do not foresee many
                     // cases where our network names include non-UTF8
                     // characters.
                     Some(s) => match s.to_str() {
-                        Some(st) => st.to_string(),
+                        Some(st) => {
+			    let name_full = st.to_string();
```
Luap99, line 77: formatting looks wrong here, I assume validate id going to fail

test/100-basic-name-resolution.bats
```diff
@@ -165,3 +165,49 @@ load helpers
...
+	expected_rc=124 run_in_container_netns "$a1_pid" "dig" "+short" "google.com" "@$gw"
+}
+
+# Internal network, but this time with IPv6. Same result as above expected.
+@test "basic container - internal network has no DNS - ipv6" {
+	setup_slirp4netns
+
+	subnet_a=$(random_subnet 6)
+	# Cloudflare and Google public anycast DNS v6 nameservers
+	create_config network_name="podman1%int" container_id=$(random_string 64) container_name="aone" subnet="$subnet_a" custom_dns_server='"2606:4700:4700::1111","2001:4860:4860::8888"' aliases='"a1", "1a"'
```
Luap99, line 198: I think you have to remove the %int from the network name and actually add the internal setting to make the tests config behave as a real config

mheon:
Fixed v4 tests. Unfortunately needed a new HashMap in the backend. Still trying to figure out why v6 is broken.

mheon:
Ah, that was easy, forgot to ask for AAAA record. Tests now passing locally.

mheon:
@Luap99 Should be ready

src/dns/coredns.rs
```diff
@@ -140,23 +140,27 @@ impl CoreDns {
                             let mut resolved_ip_list: Vec<IpAddr> = Vec::new();
                             let mut nameservers_scoped: Vec<ScopedIp> = Vec::new();
                             // Add resolvers configured for container
-                            if let Some(Some(dns_servers)) = backend.ctr_dns_server.get(&src_address.ip()) {
-                                    if !dns_servers.is_empty() {
-                                        for dns_server in dns_servers.iter() {
-                                            nameservers_scoped.push(ScopedIp::from(*dns_server));
-                                        }
+                            if backend.ctr_is_internal(&src_address.ip()) {
```
Luap99, line 143: would it make more sense to move this below to the  
"Not found, forwarding dns request for" case?

Also what is the expected behavior for a internal container, should it get NXDomain? What happens if no upstream resolvers are tried here? Do we just not answer at all and let the client timeout?

src/backend/mod.rs
```diff
@@ -120,6 +124,27 @@ impl DNSBackend {
...
+            Some(nets) => {
+                for net in nets {
+                    match self.network_is_internal.get(net) {
+                        Some(b) => is_internal &= b,
+                        None => continue,
+                    }
+                }
+            }
+            // If we don't know about you, you don't get DNS
+            None => return true,
```
Luap99, line 142: Isn't this a breaking change? I think it makes sense but AFAIK we currently allow requests from the host to do that so that can break people I think.
While I agree it would be best to have this I am not sure to include if we decide to backport

src/backend/mod.rs
```diff
@@ -120,6 +124,27 @@ impl DNSBackend {
...
+    // Checks if a container is associated with only internal networks.
+    // Returns true iff a container is only present in internal networks.
+    pub fn ctr_is_internal(&self, requester: &IpAddr) -> bool {
+        let mut is_internal = true;
+
+        match self.ip_mappings.get(requester) {
+            Some(nets) => {
+                for net in nets {
+                    match self.network_is_internal.get(net) {
+                        Some(b) => is_internal &= b,
```
Luap99, line 136: you do not have to loop over all and need the is_internal tmp var

you can check if !b { return false } and then at the bottom of the function return true.
Also rename b to something like `internal`

src/backend/mod.rs
```diff
@@ -120,6 +124,27 @@ impl DNSBackend {
         Some(results)
     }
 
+    // Checks if a container is associated with only internal networks.
+    // Returns true iff a container is only present in internal networks.
```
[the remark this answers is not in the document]
mheon: iff = if and only if (not sure how common this is though)

src/dns/coredns.rs
mheon: Internal containers should still respond correctly (assuming same network).
Unfortunately it looks like we're timing out instead of NXDOMAIN-ing. I think we can probably detect the 0 resolvers in use case and manually NXDOMAIN instead of sending it out to 0 resolvers and timing out... Let me fix that

src/backend/mod.rs
mheon: Hm. False is probably safer in that case. I don't have a strong opinion here and the risk of breaking means we should avoid this.

src/dns/coredns.rs
Luap99: there is a no_proxy case already so you can just reuse it for internal as well

Luap99 (approved):
LGTM

mheon:
Ah, tests are failing because an NXDOMAIN is still a 0 return code... Need to fix that

mheon:
Ready for merge

Luap99 (approved):
/lgtm

PR #449, opened 2024-04-10
Don't tear down all server threads on SIGHUP

### LINKED ISSUE #389: DNS requests timeout
I am using podman as a docker replacement on our gitlab-runner host. I have a 40 containers concurrency limit and when I start my tests, I get DNS resolution errors.

Testing environment:
- 16 vCPUs
- 24GB memory
- CentOS 9 stream
- podman 4.6.1-5.
- slirp4netns: slirp4netns-1.2.2-1.el9.x86_64
- aardvark-dns: aardvark-dns-1.7.0-1.el9

While running tests, I get random dns resolution fail errors inside containers (actual host replaced with host.example.tld):
```
Example 1:
Cloning into 'spec/fixtures/modules/yumrepo_core'...
ssh: Could not resolve hostname host.example.tld: Temporary failure in name resolution
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.

Example 2:
$ bundle install -j $(nproc)
Fetching gem metadata from https://host.example.tld/nexus/repository/GroupRubyGems/..
Fetching gem metadata from https://host.example.tld/nexus/repository/GroupRubyGems/..
Could not find gem 'beaker (~> 5)' in any of the gem sources listed in your

Example 3:
Initialized empty Git repository in /builds/puppet/freeradius/.git/
Created fresh repository.
fatal: unable to access 'https://host.example.tld/puppet/freeradius.git/': Could not resolve host: host.example.tld
Cleaning up project directory and file based variables
```

This does not happen in every container, it's sporadic and random. If I switch back to `cni` backend, it works without errors.

I tried running up to 8 containers and flooding the dns server with dns lookups, but I could not get a DNS resolution error. Will try to ramp that up to 30-40 and see if I can reproduce.

If anyone has an idea how to debug this, I will gladly look into it if my knowledge allows me.

kyrias:
Instead we parse the new config, replace the in-memory configuration state, stop all old threads for IPs which are no longer in the configuration, and start any new ones for IPs which were added.

Resolves #389.

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -2,27 +2,34 @@ use crate::backend::DNSBackend;
 use crate::config;
 use crate::config::constants::AARDVARK_PID_FILE;
 use crate::dns::coredns::CoreDns;
+use arc_swap::ArcSwap;
 use log::{debug, error, info};
 use signal_hook::consts::signal::SIGHUP;
 use signal_hook::iterator::Signals;
+use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fs;
+use std::hash::Hash;
 use std::net::IpAddr;
 use std::net::Ipv4Addr;
+use std::net::Ipv6Addr;
 use std::sync::Arc;
+use std::sync::OnceLock;
 use std::thread;
+use std::thread::JoinHandle;
 
-use async_broadcast::broadcast;
 use std::fs::File;
 use std::io::prelude::*;
 use std::path::Path;
 use std::process;
 
-// Will be only used by server to share backend
-// across threads
-#[derive(Clone)]
-struct DNSBackendWithArc {
-    pub backend: Arc<DNSBackend>,
-}
+type Config = (
+    DNSBackend,
+    HashMap<String, Vec<Ipv4Addr>>,
+    HashMap<String, Vec<Ipv6Addr>>,
+);
+type ThreadHandleMap<Ip> =
+    HashMap<(String, Ip), (flume::Sender<()>, JoinHandle<Result<(), std::io::Error>>)>;
 
 pub fn create_pid(config_path: &str) -> Result<(), std::io::Error> {
     // before serving write its pid to _config_path so other process can notify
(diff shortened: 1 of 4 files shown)


src/server/serve.rs
```diff
@@ -54,148 +61,179 @@ pub fn serve(
...
+        loop {
+            if signals.wait().next().is_some() {
+                break;
+            }
         }
+
+        let (new_backend, new_listen_ip_v4, new_listen_ip_v6) = parse_configs(config_path)?;
+
+        // Only replace the backend value if the configuration actually changed.
+        if **backend.load() != new_backend {
```
Luap99, line 124: My understanding is that this walks the entire map of names/ips and so on to compare if anything changed right? I guess that could get expensive with many containers, networks....
Also in general we only send SIGHUP when we actually changed something so I would suggest to not check this and always replace

Luap99 (commented):
Thanks for doing the work. I am not familiar with the async changes that you did and I a short on time so it may take a bit until I get this properly reviewed.

In the meantime did you actually test that this fixes the intermittent dns issues?

src/server/serve.rs
kyrias: Fair enough, done.

kyrias:
> Thanks for doing the work. I am not familiar with the async changes that you did and I a short on time so it may take a bit until I get this properly reviewed.

There aren't really any real async changes here to be fair, but no hurry.

Though the way aardvark is currently structured there is literally zero advantage to using any async code at all, since it's currently spawning a new tokio runtime for each thread and then essentially running a single task of straight-line code, so currently it's sort of getting the worst of both worlds.

I would argue that it should either be restructured to have a single multithreaded runtime with a task per server listened IP, or be restructured to be entirely sync.

> In the meantime did you actually test that this fixes the intermittent dns issues?

Yes, we cannot reproduce that issue with these changes.

Luap99:
> I would argue that it should either be restructured to have a single multithreaded runtime with a task per server listened IP, or be restructured to be entirely sync.

Yes agreed. Keep in mind this all was written by us when we had basically no rust experience so things not working properly here is no surprise. I think async makes sense but yeah it should be one runtime.

Luap99:
But nothing you have to do as the issues are pre-existing, I will try to get this reviewed properly in the next days

src/backend/mod.rs
```diff
@@ -8,6 +8,7 @@ use std::vec::Vec;
 // server will get back a mix of responses and filter for v4/v6 from there.
 // This may not be a good decision, not sure yet; we can split later if
 // necessary.
+#[derive(PartialEq)]
```
Luap99, line 11: I guess this can be removed

Luap99 (approved):
LGTM
@mheon @baude PTAL this will fix a annoying long standing issue with intermittent dns connectivity problems for the containers.

svenstaro:
Could we please get a release with this in it soon? Currently aardvark-dns is actually pretty broken.

mheon:
We're planning a Podman release the week of May 20th, so our next planned Netavark and Aardvark drops would coincide with that. We could do an async release earlier if this is critical.

svenstaro:
It seems to me this warrants a patch release before that seeing that this is currently really broken.

cpetestewart:
PLEASE get us a patch. I have been tearing my hair out for the last few months trying to track down this issue ever since our org had to switch from docker to podman. It was only now I found this PR that explains the problem.

cpetestewart:
Is this now in the latest podman release? When I pulled down a fresh image, it said it was updated 12 days ago.

mheon:
No. We are planning a release next week (likely wednesday) to coincide with Podman 5.1

PR #455, opened 2024-04-24
fix reverse ipv6 lookup test flake

Luap99:
The test assumed that we have a fdxx:... addresses which is what we should be getting normally, however the ipv6 subnet generator had a bug that caused it to create fdx:... sometimes which is different as ipv6 allows us to omit leading zeros, thus we actually have 0fdx which is not in the range we want to use at all.

So simply fix this by creating the proper subnet so that if the random number was less than 16 we add the leading zero add the proper place after fd.


Luap99:
@lsm5 Should we drop the RHEL 8 tasks?
`        Repository 'epel-8-x86_64' does not exist in project 'rhcontainerbot/podman-next'.`

Luap99:
@baude @mheon PTAL

PR #458, opened 2024-05-02
[skip-ci] Packit: enable rhel10, c10s tests and c10s downstream sync

lsm5:
This commit will enable downstream syncing to CentOS Stream 10. The centos maintainer will need to manually run `packit propose-downstream` and `centpkg build` until better centos integration is in place.

This commit will also help to ensure that RHEL tests are run using copr rpms built on rhel (epel) environments, while CentOS Stream tests are run using rpms on centos stream environments.

RHEL-10-Beta tests will need to depend on CentOS Stream 10 rpms until RHEL-10/EPEL-10 copr targets are available.

The `prepare` steps in the TMT tests have been rewritten to be more idiomatic and to account for CentOS Stream 10 / RHEL 10.

PR #459, opened 2024-05-06
[skip-ci] Packit: separate `packages` key for rhel jobs

lsm5:
This should prevent status reporting issues as encountered in https://github.com/containers/aardvark-dns/pull/458 . The possible cause for the error was packit getting confused between two separate copr jobs with the same parameters but different targets.

Having a separate `packages` key for rhel jobs should make it easier for packit to do the right thing.

This should also make it possible for CentOS integration tests to run automatically on all PRs and not need manual triggering by a maintainer. Those were not triggered automatically on
https://github.com/containers/aardvark-dns/pull/456 while the Fedora integration tests were.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -12,6 +12,8 @@ packages:
   aardvark-dns-centos:
     pkg_tool: centpkg
     specfile_path: rpm/aardvark-dns.spec
+  aardvark-dns-rhel:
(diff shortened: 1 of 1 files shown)


lsm5:
there's some backlog in internal testing farm, the maintainers are aware of it.

Luap99 (approved):
/lgtm

PR #461, opened 2024-05-13
[skip-ci] Packit: use `packages: [aardvark-dns-fedora] for podman-next builds

lsm5:
Otherwise multiple duplicate jobs are created on the podman-next copr.

PR #465, opened 2024-05-15
[skip-ci] TMT: Reorg and reuse upstream tests on downstream updates

lsm5:
This commit accounts for breaking changes in `dnf5 config-manager`.

Merge after https://github.com/containers/aardvark-dns/pull/461


lsm5:
@mheon @Luap99 PTAL

plans/main.fmf
```diff
@@ -30,6 +40,20 @@ adjust:
...
+          order: 40
+          script: dnf config-manager --save --setopt="*:rhcontainerbot:podman-next.priority=5"
+      because: Higher priority for podman-next packages
+
+    - when: distro >= fedora-41
+      prepare+:
+        - how: shell
+          order: 40
+          script: dnf config-manager setopt "*:rhcontainerbot:podman-next.priority=5"
+      because: Higher priority for podman-next packages. Different dnf config-manager usage for Fedora 41+
```
Luap99, line 55: Is there no if else syntax? Like this it could be possible to miss a distro accidentally, not that I think it is the case here just looking to make this more robust.

lsm5: let me check if i can simplify this further. Putting this back to draft.

lsm5: @Luap99 simplified a lot by using generic steps to work across all distros. There's still some conditionalizing for enabling centos-stream copr targets on centos-stream test environments and that can't be generalized unless there's changes in how dnf / copr enables targets on rhel/centos-stream envs.

plans/main.fmf
```diff
@@ -1,40 +1,36 @@
...
-## Ref: https://tmt.readthedocs.io/en/stable/spec/core.html#order
+prepare+:
+    - name: Install bats
+      how: shell
+      script: |
+        BATS_VERSION=1.11.0
+        curl -s -L -O https://github.com/bats-core/bats-core/archive/refs/tags/v$BATS_VERSION.tar.gz
+        tar zxf v$BATS_VERSION.tar.gz
+        cd bats-core-$BATS_VERSION
+        ./install.sh /usr
```
Luap99, line 9: Is there a strong reason to do this? I rather use the packaged version. Is the issue that there is no epel 10 yet?

edsantiago: I'm not sure what you mean by "`parallel` support". As best I understand, plain BATS will use GNU parallel just fine. What I see in your link above is that the bats rpm has a `Requires: parallel` -- that's something we could do ourselves.

For now and the immediate future this is moot, since system tests can't yet run in parallel.

lsm5: > I'm not sure what you mean by "`parallel` support". As best I understand, plain BATS will use GNU parallel just fine. What I see in your link above is that the bats rpm has a `Requires: parallel` -- that's something we could do ourselves.
> 

@edsantiago ack, I just saw `BuildRequires: parallel` in there so I thought `support`.

> For now and the immediate future this is moot, since system tests can't yet run in parallel.

In that case I would prefer what I currently have on the PR, mainly for simplicity.

Luap99: FYI we are in aardvark-dns here, we do not have parallel runs enabled but the tests are designed with sufficient isolation that it could be easily be enabled I think. 
Test timing is not that critical on this repo but I think losing this ability is bad.

Reasons why I don't like this:
- nobody ever will remember to update this there as such we keep using outdated bats forever and then don't notice if things break
- there is no trust chain, with the rpm package we get gpg signatures, sure this likely does not matter as upstream must be trusted no matter what
- there is no way to know what we get really with executing the install.sh script? What if it fails for some reason? We know have to debug this thing with no knowledge what it executes.

edsantiago: FWIW, this (fetch & unpack static tarball from github) is the method used in RHEL gating tests. It is unfortunately more reliable than epel or anything else I could think of, particularly on unreleased RHELs.

lsm5: 1 possible alternative I could try would be `dnf install $BATS_RPM_URL $PARALLEL_RPM_URL` where the URLs are generic enough that they'll work for all distros without any extra setup. I'll have to pull in parallel too because of the install requirement. I can check on that

lsm5:
I'll probably end up reworking this to also account for downstream reusability.

lsm5:
@Luap99 @edsantiago @cevich PTAL at the current state. With this setup, we can reuse the tests for Fedora zuul (before dist-git PR merge) and bodhi updates and also on CentOS Stream.

See test downstream PR: https://src.fedoraproject.org/rpms/aardvark-dns/pull-request/20 and the actual TMT test trigger in Zuul: https://artifacts.dev.testing-farm.io/34dd8354-1a3d-4d34-ba58-9fb432581b8b/ . 

Unit, validate and integration tests are run upstream while only the integration tests are run downstream.

Once everything looks ok, I'll re-enable failure notifications and repush for final merge.

test/tmt/tests.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "priority=5" >> /etc/yum.repos.d/_copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next.repo
+fi
+
+# Remove testing-farm repos if they exist because they interfere with the
+# podman-next copr. The default distro repos will not be removed and can be
+# used wherever relevant.
+rm -f /etc/yum.repos.d/tag-repository.repo
+
+# Enable EPEL on RHEL/CentOS Stream envs to fetch bats
+if [[ -n $(rpm --eval '%{?rhel}') ]]; then
```
cevich, line 39:
```suggestion
if [[ -n $RHEL_VERSION ]]; then
```

test/tmt/tests.sh
```diff
@@ -0,0 +1,52 @@
...
+    # Until EPEL 10 is available use epel-9 for all RHEL and CentOS Stream
+    dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+    sed -i 's/$releasever/9/g' /etc/yum.repos.d/epel.repo
+fi
+
+# Install dependencies for running tests
+dnf -y install bats bind-utils cargo clippy go-md2man iptables jq make netavark nftables nmap-ncat rustfmt slirp4netns
+
+cat /etc/redhat-release
```
cevich, line 48: I think this output would be more useful to have at the start of the script.  If something above goes south, execution will never reach here.

cevich (changes requested):

test/tmt/main.fmf
```diff
@@ -0,0 +1,33 @@
...
+
+/unit:
+    tag: upstream
+    summary: Unit test
+    test: make -C ../.. unit
+
+/upstream:
+    tag: upstream
+    summary: Run tests on upstream PRs
+
```
edsantiago, line 24: This and line 31 are trailing whitespace. I'm a little alarmed that the git hooks didn't block it.

edsantiago (approved):
Tentative LGTM but I'm not qualified to read packit

.packit.yaml
```diff
@@ -24,9 +37,9 @@ jobs:
   - job: copr_build
     trigger: pull_request
     packages: [aardvark-dns-fedora]
-    notifications: &copr_build_failure_notification
-      failure_comment:
-        message: "Ephemeral COPR build failed. @containers/packit-build please check."
+    #    notifications: &copr_build_failure_notification
+    #  failure_comment:
+    #    message: "Ephemeral COPR build failed. @containers/packit-build please check."
```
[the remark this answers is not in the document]
lsm5: yes

test/tmt/main.fmf
cevich: If you're talking about pre-commit hooks, I don't see any setup on this repo :disappointed:

test/tmt/tests.sh
Luap99, line 1: should this file be named test_integration.sh instead? I find tests.sh a weird name given it only runs integration tests so the purpose is not quite clear.

lsm5: sgtm

test/tmt/tests.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -exo pipefail
+
+if [[ $1 == '' ]]; then
+    echo -e "Usage: podman-tests.sh STREAM\nSTREAM can be upstream or downstream"
```
Luap99, line 6: usage script name does not match

test/tmt/main.fmf
```diff
@@ -0,0 +1,33 @@
+# Only common dependencies that are NOT required to run netavark-tests.sh are
+# specified here. Everything else is in netavark-tests.sh.
+require:
+    - cargo
+    - make
+
+adjust:
+    duration: 10m
+    when: arch == aarch64
```
Luap99, line 9: Why only on aarch?

test/tmt/tests.sh
cevich, line 6:
```suggestion
    echo -e "Usage: $(basename ${BASH_SOURCE[0]}) STREAM\nSTREAM can be upstream or downstream"
```

test/tmt/main.fmf
```diff
@@ -0,0 +1,33 @@
...
+    test: make -C ../.. unit
+
+/upstream:
+    tag: upstream
+    summary: Run tests on upstream PRs
+    
+    /integration_test:
+        test: bash tests.sh upstream
+
+/downstream:
```
Luap99, line 28: The duplication between upstream and downstream confuses me here? Also the names make no sense as this runs integration tests not upstream/downstream specific tests. 
Can't we just assign two tags upstream and downstream to one definition?

test/tmt/main.fmf
lsm5: IIRC, the default is 5m and that sufficed for x86_64 consistently. Only aarch64 timed out sometimes.

test/tmt/main.fmf
lsm5: I pushed a test change. Let me see how I can best address this. Hold off on this for a bit.

test/tmt/main.fmf
lsm5: removed

.packit.yaml
```diff
@@ -5,6 +5,19 @@
...
+    dest: gating.yaml
+    delete: true
+  - src: plans/
+    dest: plans/
+    delete: true
+  - src: .fmf/
+    dest: .fmf/
+    delete: true
+  - .packit.yaml
+
```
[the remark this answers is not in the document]
lsm5: added comment

test/tmt/main.fmf
lsm5: see [this change](https://github.com/containers/aardvark-dns/pull/465/files#diff-68ada78e2ae75611599170f255698e6e31c66e0bfe3aae153e9c41b6cb7ad89fR21-R30) which reorgs the integration tests. The results are [here](https://artifacts.dev.testing-farm.io/abce20c4-0a57-4e57-9808-82822c3e9cfa/)

Given aardvark-dns tests depend on netavark, for upstream we'll need to fetch netavark from podman-next while for downstream we'll need to fetch official fedora netavark. So, there will be some differences. We `could` move this environment detection in the script itself, but hope we're cool with doing that in a followup PR.

Luap99: yes that looks good enough to me

lsm5:
PTAL again. Good for hopefully final review and merge. Thanks!

test/tmt/tests.sh
lsm5: done

Luap99 (approved):
LGTM, I have no way to verify any of the downstream changes but looks like the upstream changes work and it is tested properly

lsm5:
> LGTM, I have no way to verify any of the downstream changes but looks like the upstream changes work and it is tested properly

Thanks @Luap99.

If anyone would like to verify how it looks like downstream (currently a manual sync), go to the dist-git test PR: https://src.fedoraproject.org/rpms/aardvark-dns/pull-request/20# -> `Zuul` CI result -> `rpm-tmt-test` -> `Artifacts` tab -> `Testing Farm Artifacts`. Only the downstream integration tests are run there.

The automatic syncing from Packit and bodhi results will only be found after an upstream release, unless we do a manual sync first. But Zuul results should be a strong enough indicator. HTH.

Luap99:
We cut a new release tomorrow should we get this merged for that?

lsm5:
> We cut a new release tomorrow should we get this merged for that?

That would be great.  Thanks @Luap99 . I have further simplifications lined up for container-selinux and netavark, which I can replicate in av as well for a followup.

PR #475, opened 2024-06-28
Be sure to have at least Epoch 2 to preserve upgrade path in c10s.

jnovy:


lsm5 (approved):
The testing-farm failures are unrelated. I'll fix those later.

/lgtm
/approve

PR #481, opened 2024-07-24
server: improve parent <-> child error handling 

Luap99:
The code currently doesn't check for most child errors as it basicaly
immediately wrote the ready byte. Also it never considered the fact that
the child can exit early and in that case it would have blocked forever
because the parent never closed the writing side before call read().

This fixes that part, however it still reports success way to early. It
should wait for the first bind but that needs more complicated changes.

serve: fix broken error logging

Turns out the task contains a result in a result and we were only
looking at task error caused by tokio and not the actual result returned
from start_dns_server().
This caused all bind error to go nowhere and no even logged by in the
journal.

Fixes: https://github.com/containers/aardvark-dns/commit/224756d98986355d7f7d71d79c8012f0c7e4c29d ("server: use only one tokio runtime")

---
Note  I still have to do more changes in order to get the bind error back to netavark on first start.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -2,13 +2,9 @@
 use crate::server::serve;
 use clap::Parser;
 use nix::unistd;
-use nix::unistd::{dup2, fork, ForkResult};
-use std::fs::File;
-use std::fs::OpenOptions;
+use nix::unistd::{fork, ForkResult};
 use std::io::Error;
-use std::io::Write;
 use std::os::unix::io::AsRawFd;
-use std::os::unix::io::FromRawFd;
 
 #[derive(Parser, Debug)]
 pub struct Run {}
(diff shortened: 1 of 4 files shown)


Luap99:
@mheon PTAL another one

src/commands/run.rs
```diff
@@ -2,13 +2,11 @@
...
 use clap::Parser;
 use nix::unistd;
-use nix::unistd::{dup2, fork, ForkResult};
-use std::fs::File;
-use std::fs::OpenOptions;
+use nix::unistd::{fork, ForkResult};
+
 use std::io::Error;
-use std::io::Write;
+
```
mheon, line 8: Any reason for the empty lines here?

mheon:
Format comment, code LGTM

src/commands/run.rs
Luap99: No just me manually deleting imports and not the whole lines
For some reason my rust integration is fine with adding missing imports but does not remove them automatically.

Luap99: I will fix this up tomorrow.

Luap99:
fix the comment but also added a new commit to switch to tokio for the signal handling.

PR #483, opened 2024-07-25
More rework and cleanups to get closer to the place where we can report bind errors

Luap99:
see commits

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -265,8 +265,7 @@ async fn start_dns_server(
     no_proxy: bool,
     nameservers: Vec<ScopedIp>,
 ) -> Result<(), anyhow::Error> {
-    let mut server = CoreDns::new(addr, port, name, backend, rx, no_proxy, nameservers)
-        .context("new dns server")?;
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon PTAL This doesn't yet include the error handling we talked about because it seems it doesn't quite work the way I hopped. I likely need to roll our own multi error type for it.

src/server/serve.rs
```diff
@@ -234,12 +207,81 @@ async fn start_dns_server(
...
-        CoreDns::new(addr, port, name, backend, rx, no_proxy).context("new dns server")?;
+    let mut server = CoreDns::new(addr, port, name, backend, rx, no_proxy, nameservers);
     server.run().await.context("run dns server")
 }
 
+async fn read_config_and_spawn(
+    config_path: &str,
+    port: u32,
+    filter_search_domain: &str,
+    handles_v4: &mut ThreadHandleMap<Ipv4Addr>,
```
mheon, line 220: So this is declared as a HashMap, but counts as a ThreadHandleMap? Fascinating.

Luap99: type inference is very good in rust
https://doc.rust-lang.org/rust-by-example/types/inference.html

PR #484, opened 2024-07-26
add tcp support

Luap99:
see commits

Listen and forward via tcp as well not just udp

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -2,6 +2,8 @@ use crate::backend::DNSBackend;
 use crate::config::constants::AARDVARK_PID_FILE;
 use crate::config::parse_configs;
 use crate::dns::coredns::CoreDns;
+use crate::error::AardvarkError;
+use crate::error::AardvarkErrorList;
 use crate::error::AardvarkResult;
 use crate::error::AardvarkWrap;
 use arc_swap::ArcSwap;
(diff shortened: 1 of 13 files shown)


src/server/serve.rs
```diff
@@ -144,16 +150,42 @@ async fn stop_and_start_threads<'a, Ip>(
...
+        let tcp_sock = match TcpListener::bind(addr).await {
+            Ok(s) => s,
+            Err(err) => {
+                errors.push(AardvarkError::wrap(
+                    format!("failed to bind tcp listener on {addr}"),
+                    err.into(),
+                ));
+                continue;
+            }
+        };
```
Luap99, line 181: @mheon here I am not sure what is best here

If we fail to bind udp or tcp we will skip that network and bind neither. I am not sure if it is worth to ignore tcp failures and still listen on udp or the other way around. I feel like this is not worth the extra complexity
WDYT?

Luap99:
looks like the bind with ncat test is flaky (I need to add some sleep there even if that sucks)

src/server/serve.rs
mheon: I think skipping the network is reasonable. Honestly, I would kind of prefer to shut down Aardvark altogether if we can't take the appropriate ports?

Luap99: Well if it is the first launch then we will exit with the right errors but if we just got reloaded via SIGHUP then we just log the error, at this point it is possible that other networks were able to bind.
Exiting means we bring down everything.

Luap99:
Ok pushed another commits to do tcp forwarding as well

TomSweeneyRedHat:
Addresses: https://issues.redhat.com/browse/RHEL-17270

PR #488, opened 2024-07-29
update upsteam resolvers on each refresh

Luap99:
Right now we read resolv.conf once at the start for each network server, this means if resolv.conf on the host changes it keeps using the old nameservers. This commit changes this to pass the same vector around for each thread. Due to rust abstractions this requires the use of a Arc<Mutex<_>> to safly share the vector between threads. Note I am using .expect() here to access the value which causes a panic on error. However if you read the docs on the mutex the only way this errors is when the thread holding the lock panicked in which case it is best for us to abort anyway.

This is still not perfect as we only refresh when a new container is started/stopped but better than not updating it at all. Long term some inotify watcher (or polling) would be better.

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -22,6 +22,7 @@ use resolv_conf::ScopedIp;
 use std::io::Error;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
+use std::sync::Mutex;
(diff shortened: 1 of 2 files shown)


Luap99:
@mheon PTAL

mheon:
LGTM

PR #489, opened 2024-07-31
config: ignore enoent errors while reading configs

Luap99:
When we list files in directory there is a race where the file might have been removed in the meantime. So ignore the ENOENT case as this is no real error and we still want to parse the other files.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1,6 +1,6 @@
 use crate::backend::DNSBackend;
 use crate::error::{AardvarkError, AardvarkResult};
-use log::warn;
+use log::error;
 use std::collections::HashMap;
 use std::fs::{metadata, read_dir, read_to_string};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
@@ -64,7 +64,19 @@ pub fn parse_configs(
                         continue;
                     }
                 }
-                let parsed_network_config = parse_config(cfg.path().as_path())?;
+                let parsed_network_config = match parse_config(cfg.path().as_path()) {
+                    Ok(c) => c,
+                    Err(e) => {
+                        if e.kind() != std::io::ErrorKind::NotFound {
+                            error!(
+                                "Error reading config file {:?} for server update: {}",
+                                cfg.path(),
+                                e
+                            )
+                        }
+                        continue;
+                    }
+                };
 
                 let mut internal = false;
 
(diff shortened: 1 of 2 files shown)


Luap99:
@mheon @baude PTAL

mheon:
LGTM

baude:
awesome
/lgtm

Luap99:
fixed the failing unit test, should be good to

Luap99:
@lsm5 look slike something is broken with testing on cetons/rhel

```
1..19
ok 1 basic container - dns itself (custom bad dns server)
ok 2 basic container - dns itself (custom good dns server)
ok 3 basic container - dns itself custom
ok 4 basic container - ndots incomplete bad entry must NXDOMAIN instead of forwarding and timing out
ok 6 basic container - dns itself with long network name
ok 7 two containers on the same network
ok 8 basic container - internal network has no DNS
ok 9 basic container - internal network has no DNS - ipv6
ok 10 two containers on different networks
ok 11 two subnets with isolated container and one shared
ok 12 three networks with a connect
ok 13 three subnets, one container on two of the subnets, network connect
ok 14 three subnets two ipaddress v6 and one ipaddress v4, one container on two of the subnets, network connect
ok 15 two containers on the same network with aliases
ok 16 check reverse lookups
ok 17 check reverse lookups on ipaddress v6
ok 18 aardvark-dns should fail when udp port is already bound
ok 19 aardvark-dns should fail when tcp port is already bound
# bats warning: Executed 18 instead of expected 19 tests
make: *** [Makefile:108: integration] Error 1
```

Look super strange, fedora seems fine

Luap99:
> ok 4 basic container - ndots incomplete bad entry must NXDOMAIN instead of forwarding and timing out
ok 6 basic container - dns itself with long network name

ok 5 is missing?! On fedora it is there.

cc @edsantiago in case you have seen this before

edsantiago:
No... BATS does not just skip tests without notification. That's really weird, and I see nothing in the test source that could explain it.

lsm5:
> ok 5 is missing?! On fedora it is there.

is this the first such occurrence ?

Luap99:
> > ok 5 is missing?! On fedora it is there.
> 
> is this the first such occurrence ?

First time but I never looked before as tmt test were broken, after https://github.com/containers/aardvark-dns/pull/465 I assumed we should have working tmt tests now. Given your PR was green I think it might be caused by recent changes in the file. I did change/add new tests in there last week.

lsm5:
> > > ok 5 is missing?! On fedora it is there.
> > 
> > 
> > is this the first such occurrence ?
> 
> First time but I never looked before as tmt test were broken, after #465 I assumed we should have working tmt tests now. Given your PR was green I think it might be caused by recent changes in the file. I did change/add new tests in there last week.

So skipping of 5 seems to be on c10s only, also seen on #491 . But there are other failures on rhel9 / c9s

PR #491, opened 2024-07-31
[skip-ci] Packit/TMT: idiomatic repo addition

lsm5:
Use idiomatic packit to configure podman-next copr repos for upstream tests and simplify test script.


lsm5:
test 5 is run on rhel-9 and c9s in this run, but being skipped on c10s.

lsm5:
@Luap99 @edsantiago any idea about the rhel9 / c9s failures?

Luap99:
> @Luap99 @edsantiago any idea about the rhel9 / c9s failures?

```
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: server used TCP
# #| expected: =~ '\(TCP\)'
# #|   actual:    '; <<>> DiG 9.16.23-RH <<>> +tcp google.com @10.45.253.1'
# #|         >    ';; global options: +cmd'
# #|         >    ';; Got answer:'
# #|         >    ';; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56432'
# #|         >    ';; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1'
# #|         >    ';; OPT PSEUDOSECTION:'
# #|         >    '; EDNS: version: 0, flags:; udp: 4096'
# #|         >    ';; QUESTION SECTION:'
# #|         >    ';google.com.			IN	A'
# #|         >    ';; ANSWER SECTION:'
# #|         >    'google.com.		179	IN	A	172.217.2.46'
# #|         >    ';; Query time: 9 msec'
# #|         >    ';; SERVER: 10.45.253.1#53(10.45.253.1)'
# #|         >    ';; WHEN: Fri Aug 02 16:18:11 UTC 2024'
# #|         >    ';; MSG SIZE  rcvd: 55'
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

the dig version doesn't output TCP in the output so the test has no way to guarantee that we actually used tcp other than the +tcp arg (which well may be good enough), on newer versions the SERVER lines contains either `(TCP)` or `(UDP)`

I like to keep the TCP output check if possible because it is easy remove the +tcp argument by accident but I don't see a good way to work around that either other than removing this check

edsantiago:
The `(TCP|UDP)` output seems to be very new, in bind 9.18 or so (it is not in bind 9.16). I'm guessing that RHEL9 has too old a version of bind-utils. @Luap99 perhaps a better safeguard might be
```bash
    assert "$output" !~ "SERVER.*UDP"
```

lsm5:
@Luap99 RE: https://github.com/containers/aardvark-dns/pull/519#issuecomment-2379676552 did you mean it worked on cs10 for you?  Going by the current test run, looks like installing bats from source behaves just the same on cs10.

lsm5:
CentOS 9 and 10 work now. PTAL.

I did see the three subnets test fail once on centos 9 stream aarch64. Flake probably?

Luap99 (approved):
/lgtm

PR #509, opened 2024-09-18
bail on broken pipe when reading udp streams

arlyon:
I have an issue with netavark + aardvark-dns + podman where under certain conditions (in my case a container running tailscale as an exit node) aardvark-dns will end up stuck burning 100% cpu (and taking down container dns resolution in the process).

When it happens, aardvark-dns continually prints log lines along the lines of `Error parsing dns message: broken pipe`. I can't see anywhere where we bail as a result so this PR attempts to resolve that. I am hoping that if we break the loop and the socket is closed then it will just reconnect and continue to work as usual.

I am still figuring out how to get this into my version of coreOS (since it is quite deeply integrated) so that I may try to repro but I wanted to open this PR in case there was any insight here (or whether this line of reasoning is sound).

Additionally, some advice on writing a test for this would be very helpful. I am trying to set up a basic rust test but it is very involved.

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -98,6 +98,7 @@ impl CoreDns {
                 },
                 v = receiver.next() => {
                     let msg_received = match v {
+                        Some(Err(err)) if err.kind() == std::io::ErrorKind::BrokenPipe => {break},
                         Some(value) => value,
                         None => {
                             // None received, nothing to process so continue
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -7,6 +7,7 @@ use futures_util::StreamExt;
 use futures_util::TryStreamExt;
 use hickory_client::{client::AsyncClient, proto::xfer::SerialMessage, rr::rdata, rr::Name};
 use hickory_proto::tcp::TcpClientStream;
+use hickory_proto::udp::DnsUdpSocket;
 use hickory_proto::{
     iocompat::AsyncIoTokioAsStd,
     op::{Message, MessageType, ResponseCode},
@@ -51,6 +52,16 @@ enum Protocol {
     Tcp,
 }
 
+trait CoreDnsUdp: DnsUdpSocket {
+    fn local_addr(&self) -> std::io::Result<SocketAddr>;
+}
+
+impl CoreDnsUdp for tokio::net::UdpSocket {
+    fn local_addr(&self) -> std::io::Result<SocketAddr> {
+        self.local_addr()
+    }
+}
+
 impl CoreDns {
     // Most of the arg can be removed in design refactor.
     // so dont create a struct for this now.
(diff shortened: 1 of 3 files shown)


src/dns/coredns.rs
```diff
@@ -87,6 +87,7 @@ impl CoreDns {
                 },
                 v = receiver.next() => {
                     let msg_received = match v {
+                        Some(Err(err)) if err.kind() == std::io::ErrorKind::BrokenPipe => {break},
```
Luap99, line 90: This doesn't look right as you break the loop for tcp and udp and there is no logic to "retry" (open the socket again). At the very least it would need to be removed from the thread_handles map so a new container spawn would allow us to add it again.

And we likely should check what kind of errors we can get here and which of those or fatal and which are recoverable in general when reading from the socket. EPIPE just doesn't make any sense to me, it is not documented in recvmsg which should be the underlying syscall (or does the tokio abstraction do anything weird)

Can you attach strace to it and provide the output of when this happens?

arlyon: I will attempt to trigger it again and see what I can do. I have attached a test as well that demonstrates the hang when a broken pipe occurs using just a simple facade.

Luap99: Without understanding why this happens it doesn't make much sense to just merge "random" fixes and hope the best. I don't see what generates EPIPE here so I like to see a strace at least to understand which syscall is returning that. I agree there might be fatal errors that should cause us to abort but this is not specific to this one I presume and not specific to udp either?

And if we do that then we must update the thread_handles correctly and I guess try to bind the socket again.

arlyon: I attached strace on my coreOS instance using a rootful toolbox container and have managed to trigger the issue again but rather stupidly did not trace child threads / processes. Trying again now.

src/dns/coredns.rs
```diff
@@ -509,3 +521,103 @@ fn reply_ip<'a>(
...
+    }
+
+    impl CoreDnsUdp for TestUdp {
+        fn local_addr(&self) -> std::io::Result<SocketAddr> {
+            Ok("127.0.0.1:0".parse().unwrap())
+        }
+    }
+
+    // we need 2 threads or tokio::spawn will block since it never yields
+    #[test_log::test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
```
Luap99, line 585: What is the point of test_log? It seem this can just be dropped so we do not need extra dependencies.

Luap99 (commented):
FYI I looked a bit around and there seem to be some errors hickory considers hard errors when reading from the socket but they do not include EPIPE https://github.com/hickory-dns/hickory-dns/commit/d2e64d8d26b3a15163ac3e176a2e40098dec8283

So I would still like to understand how you can EPIPE here.

arlyon:
Unable to repro this (and by extension provide an strace) after a few weeks of attempts so I am closing this PR. You may close the issue if you deem it necessary also

PR #507, opened 2024-09-17
[v1.10.0-rhel] cirrus.yml: update branches

Luap99:
Let's see if CI will even working here, in preparation for incoming backport request

PR #510, opened 2024-09-18
netavark_cache_groom.sh: fix wrong branch

Luap99:
As this download from the netavark repo is should use the NETAVARK_BRANCH env.

Porting from https://github.com/containers/aardvark-dns/pull/507, not failing on main as both branches are set to main here anyway but in case we have a asny branch setup in the future again we need this


Luap99:
@mheon @baude PTAL

mheon:
LGTM

Issue #508, opened 2024-09-17
Can't resolve from outside of containers.


bwskin:
With version 1.12.0 I can't get results from aardvark-dns directly from host system. Downgrade to 1.11.0 fixes it. It works correctly when accessing from container.

With v1.11.0:
```sh
~> nslookup gitea 172.20.0.1
Server:		172.20.0.1
Address:	172.20.0.1#53

Non-authoritative answer:
Name:	gitea
Address: 172.20.0.8
```
With v1.12.2, same config:
```
> nslookup gitea 172.20.0.1
Server:		172.20.0.1
Address:	172.20.0.1#53

** server can't find gitea: NXDOMAIN
```
Is that expected behaviour now? Or is it configurable (using podman config)?

Luap99:
We generally never really considered/supported the use case of resolving the names from the host directly.

I think I "broke" this in my rework (not intentionally) so if it worked before we can add it back if it is useful

bwskin:
It's useful for me, for example with ssh passing through to gitea. I just had aardvark-dns as additional resolver for some names in dnsmasq. Host can get ip address of the container to connect, and I don't have to maintain static ip for every service container i have. Pretty convenient to have it.

PR #512, opened 2024-09-19
coredns: allow host lookup of names

Luap99:
In my rework it seems I broke the weird way how we tried to resolve
names when we have no source ip match. Let's do this properly and move
the logic all into lookup() which makes it much cleaner and no
duplication needed.

Now there is catch here, we only lookup the network for the current
network listener, so if the user has multiple networks they always must
target the requests against the correct ip depeding on what network the
name is in. But this is the same as it worked before and seems safer
than allowing a unknown ip to access all names I guess.

Fixes https://github.com/containers/aardvark-dns/issues/508

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -59,7 +59,12 @@ impl DNSBackend {
 
     // Handle a single DNS lookup made by a given IP.
     // Returns all the ips for the given entry name
-    pub fn lookup(&self, requester: &IpAddr, entry: &str) -> Option<Vec<IpAddr>> {
+    pub fn lookup(
+        &self,
+        requester: &IpAddr,
+        network_name: &str,
+        entry: &str,
+    ) -> Option<Vec<IpAddr>> {
         // Normalize lookup entry to lowercase.
         let mut name = entry.to_lowercase();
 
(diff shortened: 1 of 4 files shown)


mheon:
Your tests are red.
But code looks OK. I'm a little iffy on removing the DNSResult sum type, which I think still offered some clarity, but things don't look worse without it.

Luap99:
>  I'm a little iffy on removing the DNSResult sum type, which I think still offered some clarity

Well if no caller is using the enum it adds confusion, I at least failed to understand the reason for a enum if we only ever use one option from it.

Luap99:
```
86 |             None => &vec![network_name.to_string()],
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-
   |                      |                            |
   |                      |                            temporary value is freed at the end of this statement
   |                      creates a temporary value which is freed while still in use
```

Looks like new rust versions accept this but not the older ones so I guess I have to copy here to make the old one happy.

Luap99:
@mheon @baude PTAL

mheon:
LGTM

baude:
nice simplifications here.

/lgtm

PR #513, opened 2024-09-19
[v1.10.0-rhel] Don't tear down all server threads on SIGHUP

Luap99:
Instead we parse the new config, replace the in-memory configuration state, stop all old threads for IPs which are no longer in the configuration, and start any new ones for IPs which were added.

Paul:
Fixing merge conflicts in Cargo.toml and bumping time because that no longer builds on rust v1.80

Fixes https://issues.redhat.com/browse/RHEL-59129
Fixes https://issues.redhat.com/browse/ACCELFIX-287

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,9 +7,10 @@ name = "aardvark-dns"
 version = "1.10.0"
 dependencies = [
  "anyhow",
- "async-broadcast",
+ "arc-swap",
  "chrono",
  "clap",
+ "flume",
  "futures-util",
  "hickory-client",
  "hickory-proto",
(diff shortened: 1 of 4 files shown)


Luap99:
@TomSweeneyRedHat PTAL, are these the right Jira links I added?

TomSweeneyRedHat:
@Luap99, sorry, I just saw this. Those are the right links. The RHEL one should suffice, but I prefer putting both on PR like you did.

TomSweeneyRedHat:
Not a Rust expert by any means, but the backport looks like a mirror image.

LGTM

Luap99:
@mheon PTAL

PR #515, opened 2024-09-23
Packit: disable F39 and separate out ELN

lsm5:
We have disabled Fedora 39 on podman-next copr. So, it's best to disable build and tests for them here too.

This commit also separates out ELN jobs into a separate set using `packages: [aardvark-dns-eln]` key-value.

YAML anchors are also used for reusing fedora and centos stream targets for upstream and downstream jobs.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -30,6 +30,8 @@ packages:
     specfile_path: rpm/aardvark-dns.spec
   aardvark-dns-rhel:
     specfile_path: rpm/aardvark-dns.spec
+  aardvark-dns-eln:
+    specfile_path: rpm/aardvark-dns.spec
 
 srpm_build_deps:
   - cargo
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -43,9 +45,22 @@ jobs:
...
         message: "Ephemeral COPR build failed. @containers/packit-build please check."
+    targets: &fedora_copr_targets
+      - fedora-development-x86_64
+      - fedora-development-aarch64
+      - fedora-latest-x86_64
+      - fedora-latest-aarch64
+      - fedora-latest-stable-x86_64
+      - fedora-latest-stable-aarch64
+      - fedora-40-x86_64
+      - fedora-40-aarch64
```
Luap99, line 56: Any reason to not use fedora-all here?

Luap99: I mean even if f39 is disabled in the copr it should still build fine here in the tests but I guess overall it doesn't matter and I am fine to remove it.

lsm5: mainly because of reusing `fedora_copr_targets` in the test job below where fedora-39 would fail. I'm ok to have fedora-all and only use non-fedora-39 in the tests.

PR #516, opened 2024-09-23
Release v1.10.1-rhel

Luap99:
New rhel release for
https://issues.redhat.com/browse/RHEL-59129
https://issues.redhat.com/browse/ACCELFIX-287

Also fix the cargo categories in case we want to push to crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 3
 
 [[package]]
 name = "aardvark-dns"
-version = "1.10.0"
+version = "1.10.1-rhel"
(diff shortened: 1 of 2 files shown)


Luap99:
@TomSweeneyRedHat @baude @mheon PTAL
Given this is a RHEL only release should I still push it to crates.io? I don't think it is needed for the rhel builds.

mheon:
I would not push. In the future we will probably want to move these out of the upstream repo and I don't want to make precedent regarding publishing things like this.

PR #518, opened 2024-09-25
serve: parse resolv.conf ourselves

### LINKED ISSUE #418: Setting invalid options in /etc/resolv.conf makes dns unresponsive
If you set invalid options in /etc/resolv.conf aardvark-dns will be unresponsive. It will run, but does not give any error even with RUST_LOG=trace

Note: [these options are from Oracle Solaris](https://docs.oracle.com/cd/E88353_01/html/E37852/resolv.conf-5.html) and setting these options on a RHEL-based OS will not prevent DNS requests.

This is an example file: /etc/resolv.conf
```
search this.is.dumb dont.do.this unless.you.want your.queries.to.fail like.this
options retrans:3 retry:1
nameserver 8.8.8.8
```
aardvark-dns starts:
```
RUST_LOG=trace /usr/libexec/podman/aardvark-dns --config /run/containers/networks/custom-dns -p 4343 run

ps aux
root        6587  0.0  0.0 276552   220 ?        Ssl  17:25   0:00 /usr/libexec/podman/aardvark-dns --config /run/containers/networks/backup-dns -p 4343 run
```

However, it does not respond to any queries:
```
dig @127.0.0.1 -p 4343 google.com

; <<>> DiG 9.16.23-RH <<>> @127.0.0.1 -p 4343 google.com
; (1 server found)
;; global options: +cmd
;; connection timed out; no servers could be reached
```

No errors or anything about it found in syslog:
```
cat /var/log/messages | grep dns

But shows other logs from previous testing, for example:
aardvark-dns[5506]: Unable to start server unable to start CoreDns server: Cannot assign requested address (os error 99)
aardvark-dns[5758]: Unable to start server unable to start CoreDns server: Address already in use (os error 98)
```

If you remove the bogous options from the /etc/resolv.conf file it works again. E.g. with this /etc/resolv.conf it will respond to queries as expected:
```
search this.is.dumb dont.do.this unless.you.want your.queries.to.fail like.this
nameserver 8.8.8.8
```

Tested:
aardvark-dns 1.7.0 (Podman package RHEL-based)
aardvark-dns 1.9.0 (Github Releases)

Luap99:
The resolv.conf parsing lib is super strict and does not allow unknown options, given we do not care about options search domains or really anything else in this file parse it ourselves with very lax rules.

Basically we try to find the nameserver lines. Anything else is ignored, the only error we produce is if we fail to parse the nameserver ip address.

Fixes #418
Fixes https://issues.redhat.com/browse/RHEL-57695

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,6 @@ dependencies = [
  "libc",
  "log",
  "nix",
- "resolv-conf",
  "syslog",
  "tokio",
 ]
@@ -684,12 +683,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "quick-error"
-version = "1.2.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
-
-[[package]]
 name = "quote"
 version = "1.0.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


baude:
LGTM

PR #519, opened 2024-09-27
test: make them pass on RHEL/Centos Stream 9

Luap99:
The dig version there doesn't print the protocol in the output and it is not really that important to check as we assume that dig +tcp does the right thing anyway.

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -385,7 +385,7 @@ fn parse_resolv_conf(content: &str) -> AardvarkResult<Vec<IpAddr>> {
     let mut nameservers: Vec<IpAddr> = Vec::new();
     for line in content.split('\n') {
         // split of comments
-        let line = match line.split_once(|s| s == '#' || s == ';') {
+        let line = match line.split_once(['#', ';']) {
             Some((f, _)) => f,
             None => line,
         };
--- a/test/100-basic-name-resolution.bats
+++ b/test/100-basic-name-resolution.bats
@@ -89,7 +89,9 @@ load helpers
 	run_in_container_netns "$a1_pid" "dig" "+tcp" "google.com" "@$gw"
 	# validate that we get an ipv4
 	assert "$output" =~ "[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+"
-	assert "$output" =~ "\(TCP\)" "server used TCP"
+	# TODO This is not working on rhel/centos 9 as the dig version there doesn't print the line,
+	# so we trust that dig +tcp does the right thing.
+	# assert "$output" =~ "\(TCP\)" "server used TCP"
 	# Set recursion bit is already set if requested so output must not
 	# contain unexpected warning.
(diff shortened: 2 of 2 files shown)


Luap99:
Ok this looks better, @mheon @baude @lsm5 PTAL

This leaves the weird cs10 failures
```
bats test/
1..19
ok 1 basic container - dns itself (custom bad dns server)
ok 2 basic container - dns itself (custom good dns server)
ok 3 basic container - dns itself custom
ok 4 basic container - ndots incomplete bad entry must NXDOMAIN instead of forwarding and timing out
ok 6 basic container - dns itself with long network name
ok 7 two containers on the same network
ok 8 basic container - internal network has no DNS
ok 9 basic container - internal network has no DNS - ipv6
ok 10 two containers on different networks
ok 11 two subnets with isolated container and one shared
ok 12 three networks with a connect
ok 13 three subnets, one container on two of the subnets, network connect
ok 14 three subnets two ipaddress v6 and one ipaddress v4, one container on two of the subnets, network connect
ok 15 two containers on the same network with aliases
ok 16 check reverse lookups
ok 17 check reverse lookups on ipaddress v6
ok 18 aardvark-dns should fail when udp port is already bound
ok 19 aardvark-dns should fail when tcp port is already bound
# bats warning: Executed 18 instead of expected 19 tests
make: *** [Makefile:108: integration] Error 1
```

I tried to reproduce locally with tmt but that causes even stranger errors, first of all I cannot install slirp4netns as it is removed from centos 10?! (I though that was fixed)
But even if I add it manually and run bats on the system I cannot run any test:
```
# bats test 
100-basic-name-resolution.bats
 200-two-networks.batsnternal network has no DNS - ipv6                                                                                                             9/19
 300-three-networks.batsated container and one shared                                                                                                              11/19
 400-aliases.batstwo ipaddress v6 and one ipaddress v4, one container on two of the subnets, network connect                                                       14/19
 500-reverse-lookups.batssame network with aliases                                                                                                                 15/19
 600-errors.bats lookups on ipaddress v6                                                                                                                           17/19
   aardvark-dns should fail when tcp port is already bound                                                                                                         19/19
   bats warning: Executed 0 instead of expected 19 tests

# bats --tap test 
1..19
# bats warning: Executed 0 instead of expected 19 tests
```

No idea what is wrong with bats on this system but I have to assume it is simply broken, bats version is `Bats 1.8.0` from epel repository. I guess I need to try something newer there.

Luap99:
Ok pulling bats from upstream git makes it work, no idea what is going on with the packaged version on centos stream 10 then but it is clearly not working so I guess we need to fix the install there as next step.

lsm5 (approved):
LGTM

I have another PR which removes the epel9 repo from CS10 and in that case it fetches bats from the testing-farm (fedora) repo IIRC. I'm fine with ignoring CS10 from here.

Luap99:
@lsm5 sounds good, if the repo doesn't work then I guess we should clone and install from the upstream git tag. I would love to have the packit tasks all passing here.

PR #522, opened 2024-10-08
[skip-ci] Packit: constrain koji job to the fedora package

lsm5:
This helps to avoid dup downstream jobs. Doesn't affect upstream.


Luap99 (approved):
/lgtm

PR #523, opened 2024-10-09
coredns: forward names with no ndots as well

Luap99:
Docker does this and there might be local names and a local resolver that otherwise cannot be resolved.

Fixes https://issues.redhat.com/browse/RHEL-61873

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -202,7 +202,6 @@ impl CoreDns {
         if data.no_proxy
             || backend.ctr_is_internal(&src_address.ip())
             || request_name_string.ends_with(&backend.search_domain)
-            || request_name_string.matches('.').count() == 1
         {
             let mut nx_message = req.clone();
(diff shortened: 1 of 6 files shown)


baude:
yeah we discussed this ... LGTM @mheon ping

mheon:
LGTM, but your tests are red

Luap99:
yes tests need more work, some requests time out now instead of the nxdomain

One thing I realized instead of using slirp4netns and connecting to public dns servers maybe I should do the dnsmasq setup for each test with a few dummy records we can query. This solves our network dependency and some other issues, i.e. I can add big dns entries that must use tcp and cannot be resolved via udp which gives better coverage overall.
But I put this below all the other prority work so far so I didn't get to it yet.

Luap99:
@mheon @baude PTAL again, now with a bunch of test changes that I think are better as I removed external network access from the tests

PR #525, opened 2024-10-15
define a msrv policy

Luap99:
like in https://github.com/containers/netavark/pull/1098


Luap99:
@baude @mheon PTAL

mheon:
LGTM

PR #530, opened 2024-10-28
run cargo update

Luap99:


Luap99:
@mheon @baude PTAL

tmt tests fail with 

>  {"error":"unable to append rule '-j MARK  --set-xmark 0x2000/0x2000' to table 'nat': code: 2, msg: Warning: Extension MARK revision 0 not supported, missing kernel module?\nip6tables v1.8.10 (legacy): unknown option \"--set-xmark\"\nTry `ip6tables -h' or 'ip6tables --help' for more information.\n"}

This is a known kernel bug https://github.com/containers/podman/issues/24374 so it can be ignored, the cirrus tests pass

PR #531, opened 2024-10-29
Release v1.13.0

baude:
* Set TTL to 0 for container names
* Allow forwarding of names with no ndots
* DNS: limit to 3 resolvers and use better timeout for them
* Ignore unknown resolv.conf options


Luap99 (approved):
LGTM
@mheon PTAL

PR #532, opened 2024-10-29
tmt: install dnsmasq dependency

Luap99:
We need this as of commit e6ce710660.

### DIFF
--- a/test/tmt/test_integration.sh
+++ b/test/tmt/test_integration.sh
@@ -46,7 +46,7 @@ if [[ -n $RHEL_VERSION ]]; then
 fi
 
 # Install dependencies for running tests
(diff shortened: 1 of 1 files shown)


Luap99:
Ok this makes the fedora tests work again.

lsm5 (approved):
LGTM

I'll remove the rhel tests from aardvark-dns as well. If we're not testing on netavark, no need here either.

Luap99:
@lsm5 please merge this, there really isn't much of a point forcing a second review on such small non code changes.

Luap99:
nevermind I see you already added this in https://github.com/containers/aardvark-dns/pull/491

PR #533, opened 2024-10-31
[skip-ci] Packit: Remove epel jobs

lsm5:
epel environments on testing farm and copr envs (RHEL+EPEL) are often out of date causing build and testing issues.

We're testing continuously on centos stream so we can ensure that whatever ultimately enters RHEL has been tested upstream.


lsm5:
handled in #491

PR #534, opened 2024-10-31
[skip-ci] Packit/TMT: idiomatic repo addition

lsm5:
Switch to idiomatic repo addition to test environments instead of adding them using dnf commands.
    
EPEL-10 is now available, so epel repo addition to centos stream 10 repos should just work.


lsm5:
duplicate of #491

PR #539, opened 2024-11-05
test: make jq query work on centos stream 9

Luap99:
The syntax is not accepted by older version like on centos stream 9, this one should work with all versions and thus make the tmt tests on centos 9 stream pass again.


Luap99:
@lsm5 PTAL, looks like it was simple enough to get it working again

lsm5 (approved):
LGTM. thanks a lot!

PR #541, opened 2024-11-06
Release 1.13.1

Luap99:


Luap99:
@mheon @baude PTAL
I didn't add a -dev bump here as I don't think we need one on a release branch but I can add it if you prefer having that.

mheon:
LGTM

PR #557, opened 2024-12-24
[skip-ci] RPM/TMT: use tests subpackage

lsm5:
This makes fetching and running of gating tests similar to podman, buildah and skopeo.

### DIFF
--- a/rpm/aardvark-dns.spec
+++ b/rpm/aardvark-dns.spec
@@ -78,6 +78,10 @@ tar fx %{SOURCE1}
 %install
 %{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install
 
+# Add empty check section to silence rpmlint warning.
+# No tests meant to be run here.
+%check
+
 %files
 %license LICENSE
 %if (0%{?fedora} || 0%{?rhel} >= 10) && !%{defined copr_username}
--- a/rpm/aardvark-dns.spec
+++ b/rpm/aardvark-dns.spec
@@ -53,6 +53,23 @@ BuildRequires: rust-srpm-macros
 Forwards other request to configured resolvers.
 Read more about configuration in `src/backend/mod.rs`.
 
+%package tests
+Summary: Tests for %{name}
+
+Requires: %{name} = %{epoch}:%{version}-%{release}
+Requires: bats
+Requires: bind-utils
+Requires: jq
+Requires: netavark
+Requires: nmap-ncat
+Requires: dnsmasq
+
+%description tests
+%{summary}
+
+This package contains system tests for %{name} and is only intended to be used
+for gating tests.
+
 %prep
 %autosetup -Sgit %{name}-%{version}
 # Following steps are only required on environments like koji which have no
@@ -78,6 +95,10 @@ tar fx %{SOURCE1}
 %install
 %{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install
 
+%{__install} -d -p %{buildroot}%{_datadir}/%{name}/test
+%{__cp} -rp test/* %{buildroot}%{_datadir}/%{name}/test/
+%{__rm} -rf %{buildroot}%{_datadir}/%{name}/test/tmt/    
+
 # Add empty check section to silence rpmlint warning.
 # No tests meant to be run here.
 %check
@@ -90,6 +111,9 @@ tar fx %{SOURCE1}
 %endif
 %dir %{_libexecdir}/podman
 %{_libexecdir}/podman/%{name}
+
+%files tests
+%{_datadir}/%{name}/test
 
 %changelog
 %autochangelog
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,6 +16,11 @@ files_to_sync:
   - src: plans/
     dest: plans/
     delete: true
+    mkpath: true
+  - src: test/tmt
+    dest: test/tmt
+    delete: true
+    mkpath: true
   - src: .fmf/
     dest: .fmf/
     delete: true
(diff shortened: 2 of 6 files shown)


rpm/aardvark-dns.spec
```diff
@@ -53,6 +59,31 @@ BuildRequires: rust-srpm-macros
 Forwards other request to configured resolvers.
 Read more about configuration in `src/backend/mod.rs`.
 
+# Only intended to be used for gating tests
+# End user usecases not supported
```
Luap99, line 63: should this be part of the description? So end user have actually a chance to read it.

rpm/aardvark-dns.spec
```diff
@@ -53,6 +59,31 @@ BuildRequires: rust-srpm-macros
...
+Requires: bats
+%else
+Recommends: bats
+%endif
+Requires: bind-utils
+Requires: iptables
+Requires: jq
+Requires: make
+Requires: netavark
+Requires: nftables
```
Luap99, line 78: Isn't nftables already required by netavark?

rpm/aardvark-dns.spec
```diff
@@ -53,6 +59,31 @@ BuildRequires: rust-srpm-macros
...
+Requires: %{name} = %{epoch}:%{version}-%{release}
+%if %{defined bats_ofc}
+Requires: bats
+%else
+Recommends: bats
+%endif
+Requires: bind-utils
+Requires: iptables
+Requires: jq
+Requires: make
```
Luap99, line 76: make doesn't seem to be needed to run as your script just calls `bats test/`

Luap99 (commented):
Overall seems fine, I cannot really comment on the tmt/gating specifics but having a test subpackage to run them in gating seems good.

lsm5:
@Luap99 cs9 and f40 tests are failing on not finding nft. Are the tests there expected to fail without nftables? IIUC, the default_fw is nftables only on f41+ and cs10.

Luap99:
> @Luap99 cs9 and f40 tests are failing on not finding nft. Are the tests there expected to fail without nftables? IIUC, the default_fw is nftables only on f41+ and cs10.

Sorry missed the ping earlier, I think https://github.com/containers/aardvark-dns/pull/568 should fix it.

lsm5:
@Luap99 @mheon PTAL. #568 does fix issues. Thanks.

rpm/aardvark-dns.spec
```diff
@@ -53,6 +53,24 @@ BuildRequires: rust-srpm-macros
 Forwards other request to configured resolvers.
 Read more about configuration in `src/backend/mod.rs`.
 
+%package tests
+Summary: Tests for %{name}
+
+Requires: %{name} = %{epoch}:%{version}-%{release}
+Requires: bats
+Requires: bind-utils
+Requires: iptables
```
Luap99, line 62: the iptables should be dropped, the depency is set by netavark already

rpm/aardvark-dns.spec
```diff
@@ -78,6 +96,14 @@ tar fx %{SOURCE1}
 %install
 %{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install
 
+%{__install} -d -p %{buildroot}%{_datadir}/%{name}/test
+%{__cp} -rp test/* %{buildroot}%{_datadir}/%{name}/test/
+%{__rm} -rf %{buildroot}%{_datadir}/%{name}/test/tmt/    
+
+# Add empty check section to silence rpmlint warning.
+# No tests meant to be run here.
+%check
```
Luap99, line 105: Looking around online this seems to say


> Command or series of commands to test the software. This normally includes things such as unit tests.

https://rpm-packaging-guide.github.io/

So should we run unit tests there? I see you do have the unit test listed under the actual tests which I guess only run upstream? I have experience around that here so any reason not to run the unit test at least part of the normal package build?

Luap99 (commented):
Looking at the testing farm tasks they only run integration tests on this PR?
From you commit on latest fedora we should run all validate and unit and integration. I don't see that looking at f42 and f41 and f40 tasks so I don't think it is working

rpm/aardvark-dns.spec
lsm5: Do the unit tests depend on the generated aardvark-dns binary at all? If not, I don't see any point in running them in `%check` .

rpm/aardvark-dns.spec
lsm5: ack. Updating.

lsm5:
> Looking at the testing farm tasks they only run integration tests on this PR? From you commit on latest fedora we should run all validate and unit and integration. I don't see that looking at f42 and f41 and f40 tasks so I don't think it is working

Added unit and validate tests back into the same packit job, so you'll see them together now and they're enabled for all targets.

rpm/aardvark-dns.spec
Luap99: Unit test do not depend on the build binary in any way. I don't mind either way we run them and using tmt is perfectly fine

Luap99: Looking at the rpm build log that does indeed run the unit tests as part of the rpm build so I guess if unit tests fail we do not get a rpm. Whenever that is preferable or not I cannot judge and leave it to you if you want that.

The unit tests here are not particular distro depended so I think it is  fine if we only run it upstream

lsm5: ack thanks. unit test failures will be visible to us upstream anyway, so I'd say let's not keep them as additional blockers for the rpm. And not blocking on unit tests for rpm has worked mostly fine for us so far AFAICT.

Removing it from %check.

lsm5:
@Luap99 @mheon PTAL. I think this should be good to go.

Luap99 (approved):
LGTM

lsm5:
@Luap99 @mheon @flouthoc this one should be good for merge.

PR #562, opened 2025-01-19
Add vrf compatibility

MrSuicideParrot:
Netavark  [version v1.8.0](https://github.com/containers/netavark/releases/tag/v1.8.0) added support for bridges inside VRFs.

Currently, aardvark-dns  doesn't work on bridges with a configured VRF. Aardvark-dns attempts to bind to the bidge's IP. However, as the IP is assigned to a VRF, it fails to bind. For a program to be able to bind to a VRF, it must first [bind the socket to the VRF where the IP exists](https://docs.kernel.org/networking/vrf.html). This is done using the  `SO_BINDTODEVICE` socket option.

This pull request adds this capability: before binding to an IP address, it checks if the  bridge interface is assigned to a VRF, and if so, it binds the socket to the VRF before attempting to bind the socket to the IP address. 

Unfortunately, this feature significantly complicates the logic of binding the DNS server and  forces us to use lower-level APIs to set this option. Moreover, to check if a bridge is set to a VRF, I used methods from netavark, thus increasing the number of dependencies of this project.

For these reasons, I don't know if there is interest in merging this pull request, but I did it for at least if anyone encounters this problem and wants it solved, there is already a solution.


Cargo.toml
```diff
@@ -35,6 +35,9 @@ nix = { version = "0.29.0", features = ["fs", "signal", "net"] }
 libc = "0.2.168"
 arc-swap = "1.7.1"
 flume = "0.11.1"
+netavark = "1.13.1"
```
Luap99, line 38: We cannot do this here. This dependency would make it very hard for us to cut new releases.

src/utils/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        },
+        None => {
+            return None;
+        }
+    }
+
+    match vrf_index {
+        Some(vrf_index) => {
+            let vrf_link_msg = host.get_link(netlink::LinkID::ID(vrf_index)).unwrap();
```
Luap99, line 44: in general using unwrap() is almost never acceptable, error must be handled properly

Luap99 (commented):
This seems backwards, netavark has already all the info we need so there should be no need at all to lookup any addresses via netlink to figure out if there is a vrf.

I have not looked at how vrf works here and if we really must bind the device instead but if extra info is needed then they must be part of the aardvark-dns config file we read instead.

MrSuicideParrot:
I have tried to bind to the vrf  without this option but I haven't had success. There is a tunable kernel option that enables programs to bind to any vrf, but this is only applicable when a program bind to all interfaces (0.0.0.0) and not a single IP as aardvark-dns does.

I understand the problem of adding the netavark as a depency. I'll investigate 
how can I add the vrf information to the config file and remove this dependency.

Luap99:
you need to extend the config file format https://github.com/containers/aardvark-dns/blob/main/config.md#first-line

I guess we could just add the device name on the first line. Then have netavark set this to the vrf name and call bind to device.

PR #576, opened 2025-03-07
.github: remove cirrus rerun action

Luap99:
As pointed out in buildah[1] the action is broken in bad ways where it can trigger 1000+ rerun wasting our cloud resources.

We have little flakes here and in general blindly re-rerunning is not the best so just get rid of it for now until we find something better.

[1] containers/buildah#6035


Luap99:
@baude @mheon PTAL and merge

mheon:
LGTM

PR #581, opened 2025-03-13
test: use ncat not nc

Luap99:
nc can be provided by either ncat (nmap) or netcat (OpenBSD), we only work with the nmap version so make sure we always use that one and not the short alias which can be resolved to either one.

It is not clear to me what changed on rawhide but it seems on the tmt env nc is preferred even though we have nmap-ncat installed.

### DIFF
--- a/test/100-basic-name-resolution.bats
+++ b/test/100-basic-name-resolution.bats
@@ -75,7 +75,7 @@ function teardown() {
 	setup_dnsmasq
 
 	# using sh-exec to keep the udp query hanging for at least 3 seconds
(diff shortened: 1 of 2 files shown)


Luap99:
This should make the tmt tests on rawhide pass again, i.e. see errors from https://artifacts.dev.testing-farm.io/6b99b785-1e57-42aa-af42-a5251b0e20fa/

cc @lsm5

Luap99:
@mheon @baude PTAL

flouthoc (approved):
LGTM

PR #596, opened 2025-05-02
deps: update nix to 0.30.1

Luap99:
The dup2 API has changed and now requires an OwnedFd as traget fd, that is not what we have or want. An OwnedFd is closed on drop so even if I would create one it would be dropped at the end of the function which would then close stdin/out/err which would be clearly wrong.

To accommodate the common use case of replacing the std stream nix added the dup2_stdin/stdout/stderr functions instead which we now use.

Also read no longer accepts a raw fd either.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,9 +759,9 @@ dependencies = [
 
 [[package]]
 name = "nix"
-version = "0.29.0"
+version = "0.30.1"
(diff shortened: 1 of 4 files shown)


baude:
LGTM

Luap99:
@mheon PTAL

PR #597, opened 2025-05-02
ci: ensure build does not modify committed files

Luap99:
ci: ensure tree is clean after build

To ensure Cargo.lock or other files that are committed are not modified
by the build in CI.

---

Makefile: use cargo --locked in CI

CI should never modify the lockfile so add --locked when running under
CI, the CI env is always defined by cirrus.

https://doc.rust-lang.org/cargo/commands/cargo.html#manifest-options

(like the netavark PR https://github.com/containers/netavark/pull/1233)


Luap99:
@mheon PTAL

PR #607, opened 2025-06-03
make more use of AardvarkError

Luap99:
A clippy lint[1] wants us to use the other() function on the io error. However we already have our own error type for custom messages so just use that and no co-opt the std::io::Error type.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#io_other_error


baude:
two lgtms

/lgtm

Issue #605, opened 2025-05-30
aardvark-dns only serves one DNS request per TCP connection


shenghaoyang:
Hi folks,

aardvark-dns currently only serves one DNS request per TCP connection before dropping the `TcpStream` and closing it - there's no loop around https://github.com/containers/aardvark-dns/blob/abf96b2638a0d94283beddf44719d7c811394a6c/src/dns/coredns.rs#L122

This fails name resolution for certain records using the glibc resolver:

```
podman network create testing
podman run --network testing --rm quay.io/fedora/fedora:42 getent ahosts login.microsoftonline.com
<EMPTY>
```

We've added a loop to handle the second request in our fork but can't submit a patch due to company policy.

Thanks!

Luap99:
Your reproducer works fine for me and it shows the resolved ip addresses.

But I get your point about not being able to reuse the same tcp connection.

> We've added a loop to handle the second request in our fork but can't submit a patch due to company policy.

A PR would be welcome of course as I don't have time to work on this at the moment. Any chance you can work something out?

Luap99:
The reason I avoided the loop originally is because I didn't want to reuse the TCP stream to avoid client opening many long lived connections that may never get closed.

rfc7766 mentions rate limiting the open connections by ip which seems reasonable but more complicated to implement and also to close the connections after some timeout which we already have at 3s.

shenghaoyang:
> The reason I avoided the loop originally is because I didn't want to reuse the TCP stream to avoid client opening many long lived connections that may never get closed.
> 
> rfc7766 mentions rate limiting the open connections by ip which seems reasonable but more complicated to implement and also to close the connections after some timeout which we already have at 3s.
> 

Yeah that's understandable - hopefully an idle timeout of 3s (which the loop conversion will turn this into) would be acceptable? We could also hoist that timeout up and wrap the entire loop in it.

I'm not very sure why that case fails - we _do_ have quite a few search domains defined in resolv.conf. Maybe that's the difference?

We did try looking into glibc's resolver implementation and it appears to place some emphasis on the success of that second AAAA query, but we've not figured out why.

Not replying triggers searches which NXDOMAIN.

What's more interesting is that the query doesn't fail for  other common domains like `google.com`. 

We tried other clients as well: `dig` does just fine, and so does the `musl` resolver.

Luap99:
[...]

PR #614, opened 2025-07-17
coredns: process more than one message per tcp connection

Luap99:
DNS over TCP allows for multiple requests per connection. Right now we only processed the first one which seems to cause problems for the glibc resolver.

Fix this by allowing multiple requests, however we still keep our 3s timeout per message so if the connection is unused for longer we still close it to avoid keep to many connections open at the same time.

Fixes: #605
Fixes: https://issues.redhat.com/browse/RHEL-104209

## Summary by Sourcery

Enable processing of multiple DNS messages over a single TCP connection while retaining a per-message timeout to prevent idle connections from lingering.

New Features:
- Allow CoreDNS to handle multiple DNS-over-TCP requests on the same connection

Enhancements:
- Retain a 3-second timeout for each TCP message to avoid excessive open connections

Tests:
- Add a bats test using dig +keepopen to verify multiple TCP queries are served correctly

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -130,7 +130,7 @@ impl CoreDns {
                 },
                 Err(_) => {
                     debug!(
-                        "Tcp connection {peer} was cancelled after 3s as it took to long to receive message"
+                        "Tcp connection {peer} was cancelled after 3s as it took too long to receive message"
                     );
                     break;
                 }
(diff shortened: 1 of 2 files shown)


Luap99:
@mheon @flouthoc PTAL

mheon:
LGTM

PR #619, opened 2025-08-09
test: Remove empty user= directive in dnsmasq.conf to avoid SIGSEGV

### LINKED ISSUE #618: Some tests fail with dnsmasq dumping core
Some tests fail with dnsmasq dumping core.  Upon closer inspection, the culprit seems to be the emtpy `user=` directive used in [dnsmasq.conf](https://github.com/containers/aardvark-dns/blob/main/test/dnsmasq.conf).  This may be a bug in dnsmasq.

Tracking in SUSE as https://bugzilla.opensuse.org/show_bug.cgi?id=1247812

openSUSE Tumbleweed 20250806
aardvark-dns: aardvark-dns-1.15.0-1.1
dnsmasq version: dnsmasq-2.91-1.1

The latest jobs can be found at:
- aarch64: https://openqa.opensuse.org/tests/latest?distri=opensuse&flavor=DVD&version=Tumbleweed&arch=aarch64&test=container_host_aardvark_testsuite
- x86_64: https://openqa.opensuse.org/tests/latest?distri=opensuse&flavor=DVD&version=Tumbleweed&arch=x86_64&test=container_host_aardvark_testsuite

The core files can be found in the `Logs & Assets` tab.

ricardobranco777:
Remove empty user= directive in dnsmasq.conf to avoid SIGSEGV

Fixes https://github.com/containers/aardvark-dns/issues/618

Verification runs (openSUSE Tumbleweed):
- aarch64: https://openqa.opensuse.org/tests/5231401
- x86_64: https://openqa.opensuse.org/tests/5231402

All previously ignored tests (100-basic-name-resolution 200-two-networks 300-three-networks) are now enabled in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/22928

## Summary by Sourcery

Remove the empty user= directive from the test dnsmasq configuration to prevent segfaults when running aardvark-dns, as reported in issue #618.

Bug Fixes:
- Remove empty user= line in dnsmasq.conf to avoid SIGSEGV

Tests:
- Verify no core dumps are generated
- Confirm 200-two-networks and 300-three-networks tests now pass

### DIFF
--- a/test/dnsmasq.conf
+++ b/test/dnsmasq.conf
@@ -6,8 +6,6 @@ no-resolv
 
 log-queries
 
-user=
-
 # aone and bone should return NXDOMAIN, by default dnsmasq returns REFUSED
 address=/aone/
 address=/bone/


Luap99 (approved):
LGTM @mheon @flouthoc PTAL

I think I added to avoid it switching the user to dnsmasq and just run as root but given it works without that fine this seems good.

That said I suggest you report this against dnsmasq, at the very least parsing a config file should never SIGSEGV I would say.

ricardobranco777:
> I think I added to avoid it switching the user to dnsmasq and just run as root but given it works without that fine this seems good.
> 
> That said I suggest you report this against dnsmasq, at the very least parsing a config file should never SIGSEGV I would say.

Yes.  I contacted the maintainer via e-mail.

flouthoc (approved):
LGTM

Issue #622, opened 2025-08-14
Drop dependency on ncat as test helper


ricardobranco777:
I've seen the tests work with very specific versions of ncat like 7.92 & 7.95 while 7.94 & 7.97 fail.

I'm afraid we need a test helper like the one done for netavark in https://github.com/containers/netavark/pull/1287

I'd go even further and make this a separate tool to make it available for podman too.

Luap99:
> But podman is still using ncat in other tests like 252-quadlet & 500-networking.

They only use it to bind a port/send data, we could replace that with socat as well.

PR #623, opened 2025-08-14
test: replace ncat with socat

Luap99:
ncat keeps causing issues with different versions having different
behaviors. And many distro ship different incompatible versions so it is
hard to make the test work everywhere consistently.

To address replace ncat with socat. Socat seems to much more stable and
we have been using it in podman for pasta tests for a while without
problems.

Fixes: #622

### DIFF
--- a/rpm/aardvark-dns.spec
+++ b/rpm/aardvark-dns.spec
@@ -61,7 +61,7 @@ Requires: bats
 Requires: bind-utils
 Requires: jq
 Requires: netavark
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon @baude @flouthoc PTAL

flouthoc (approved):
LGTM

ricardobranco777:
I could backport this beauty even to v1.12.2.  Thanks!

PR #628, opened 2025-09-05
Update nameservers when resolv.conf is modified.

ashley-cui:
When resolv.conf is modified, aardvark-dns now re-reads resolv.conf, and updates the nameservers accordingly.

## Summary by Sourcery

Monitor /etc/resolv.conf for changes and automatically reload upstream nameservers without restarting the server.

New Features:
- Watch /etc/resolv.conf via inotify and dynamically update the in-memory nameserver list on modification

Build:
- Add inotify crate dependency

Tests:
- Add integration test to verify resolver behavior switches after updating resolv.conf

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -26,10 +26,14 @@ use tokio::net::{TcpListener, UdpSocket};
 use tokio::signal::unix::{signal, SignalKind};
 use tokio::task::JoinHandle;
 
+use futures::StreamExt;
+use inotify::{EventStream, Inotify, WatchMask};
 use std::fs::File;
 use std::io::prelude::*;
 use std::path::Path;
 use std::process;
+
+const RESOLV_CONF: &str = "/etc/resolv.conf";
 
 type ThreadHandleMap<Ip> =
     HashMap<(String, Ip), (flume::Sender<()>, JoinHandle<AardvarkResult<()>>)>;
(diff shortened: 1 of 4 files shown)


src/server/serve.rs
```diff
@@ -176,7 +178,7 @@ where
                 continue;
             }
         };
-
+        tokio::spawn(monitor_upstream_resolvers(nameservers.clone()));
```
Luap99, line 181: This is the wrong place to do this, essentially you create a task leak because every time aardvark-dns gets reloaded (i.e. each container start/stop) you create a new task but the older one never terminates.
this would need to happen in serve() before the loop to only start it once.

src/server/serve.rs
```diff
@@ -435,6 +437,50 @@ fn parse_resolv_conf(content: &str) -> AardvarkResult<Vec<SocketAddr>> {
     Ok(nameservers)
 }
 
+async fn monitor_upstream_resolvers(nameservers: Arc<Mutex<Vec<SocketAddr>>>) {
+    let mut inotify = match Inotify::init() {
+        Ok(inotify) => inotify,
+        Err(err) => {
+            error!("Failed to initialize inotify: {err}");
+            warn!("Nameservers will not be updated on resolv.conf change");
+            return;
```
Luap99, line 446: Properly best to log this as one line.

Though from experience inotify can fail so I wonder if we should also implement a timer based polling fallback, i.e. read the resolv.conf every 5s or so.

src/server/serve.rs
```diff
@@ -435,6 +437,50 @@ fn parse_resolv_conf(content: &str) -> AardvarkResult<Vec<SocketAddr>> {
...
+        }
+    };
+    if let Err(err) = inotify
+        .watches()
+        .add(
+            "/etc/resolv.conf",
+            WatchMask::MODIFY,
```
Luap99, line 453: it makes sense to put `/etc/resolv.conf` into a const and then use it here and in get_upstream_resolvers()

And I don't think we want to listen on modify, if I look into dnsmasq it calls
`inotify_add_watch(daemon->inotifyfd, path, IN_CLOSE_WRITE | IN_MOVED_TO);` which make sense. We would not want to try to read partial changes if we can avoid it only when the file is closed. And I assume we also must handle th emove case because in most cases to be "atomic" most implementations that chnage the file with actually create a tmpfile and then rename it instead.

src/server/serve.rs
```diff
@@ -435,6 +437,50 @@ fn parse_resolv_conf(content: &str) -> AardvarkResult<Vec<SocketAddr>> {
...
+        ) {
+        error!("Failed to add watch on /etc/resolv.conf: {err}");
+        warn!("Nameservers will not be updated on resolv.conf change");
+        return;
+    }
+        let mut buffer = [0; 1024];
+        loop{
+            let events = match inotify.read_events_blocking(&mut buffer) {
```
Luap99, line 461: it seems inotify has already an async API for tokio so it would be better if we can use that to avoid spawning a blocking task. 
https://docs.rs/inotify/latest/inotify/struct.EventStream.html

note you must add the `stream` feature to the Cargo.toml for this dep.

basically create the eventstream in serve() and the in the main loop where we call `signals.recv().await` that should be switched to a `tokio::select! {}` where we can wait for both the signal and inotify events in "parallel"

src/server/serve.rs
```diff
@@ -88,27 +92,72 @@ pub async fn serve(
...
+                    error!("{e}");
+                };
+            }
+            // Block until resolv.conf is changed, monitored via inotify. Then reload nameservers
+            _ = async {
+                if let Some(ref mut stream) = event_stream {
+                    stream.next().await
+                } else {
+                    // If inotify setup failed, then block this branch of tokio select forever
+                    futures::future::pending().await
```
ashley-cui, line 140: Not sure if this is better or if I should just have the tokio select in an if statement and have some code dup.

ashley-cui:
@Luap99 Thanks for your patient reviews of my newbie rust code!! PTAnotherL

PR #663, opened 2026-02-11
Cargo.toml: bump indirect time dep to 0.3.47

Luap99:
Due to cve-2026-25727, we have not yet done an analysis if we are actually affected but the bump will not hurt for sure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,9 +869,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
 
 [[package]]
 name = "num-traits"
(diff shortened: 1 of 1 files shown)


lsm5:
```
$ cargo tree -p time --format "{p} {f}"
time v0.3.44 alloc,default,formatting,local-offset,std
```
So, `parsing` feature isn't enabled and the vulnerable code path is under `parsing`. So, AFAICT we're not affected.

lsm5 (approved):
LGTM

Luap99:
> ```
> $ cargo tree -p time --format "{p} {f}"
> time v0.3.44 alloc,default,formatting,local-offset,std
> ```
> 
> So, `parsing` feature isn't enabled and the vulnerable code path is under `parsing`. So, AFAICT we're not affected.

Ack thanks for checking that makes sense, it would have surprised me if something is parsing timestamps here

Luap99:
forced pushed to reset the CI as the outage seem to have left the job in a bad spot where rerun did not work

danishprakash:
> https://github.com/advisories/GHSA-r6v5-fh4h-64xc, we have not yet done an analysis if we are actually affected

I spent some time on this, and it seems that aardvark-dns is _not_ affected; the vuln exists in the RFC 2822 format parser, and the two consumers of time in aardvark-dns don't use RFC 2822 format for parsing the time(utc_now), let alone external strings.

Luap99:
This should be good to merge now since we bumped to msrv 1.88

mheon (approved):
LGTM

PR #692, opened 2026-04-30
Packit: Only create dist-git PRs for rawhide

lsm5:
We'll only be releasing breaking changes on rawhide so dist-git PRs should only be created for rawhide.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -123,12 +123,16 @@ jobs:
             id: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/repo/centos-stream-$releasever/rhcontainerbot-podman-next-centos-stream-$releasever.repo
 
   # Sync to Fedora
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -123,12 +123,16 @@ jobs:
...
+  # Rawhide (to be Fedora 45) and newer.
+  # TODO: Update dist_git_branches as and when new Fedora releases are
+  # available.
   - job: propose_downstream
     trigger: release
     packages: [aardvark-dns-fedora]
     update_release: false
     dist_git_branches: &fedora_targets
-      - fedora-all
+      - fedora-rawhide
```
Luap99, line 135: I assume this isn't an option in packit but it would be cool if we could have fedora > 44 here somehow to not have to update this later again.

lsm5: ya, nothing so far. Those are all we can use: https://packit.dev/docs/configuration#aliases

Luap99 (approved):
LGTM

PR #695, opened 2026-05-08
bump hickory to 0.26

Luap99:
Contains several breaking changes.

PR #697, opened 2026-05-08
No cirrus

lsm5:
See individual commits.


Luap99:
Thanks. Note to myself once this is ready we need to update merge protection settings with the new job names.

I think we looked at this a while ago but can packit do do task dependencies? I suppose it is not critical but it is nice to first run validate before triggering all the other jobs.
And that would allow us to have final status task, like "Total success". Without this in the github merge protection settings I need to set all exact task names like show in the github interface, `testing-farm:fedora-44-x86_64:validate`, etc... which is highly annoying because not only do I need to set a ton of names these names will change with distro upgrades so require ongoing churn for updates.

test/tmt/main.fmf
```diff
@@ -24,3 +27,17 @@
     adjust:
         duration: 10m
         when: arch == aarch64
+
+/msrv:
+    summary: Minimum Supported Rust Version (MSRV) check
+    test: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.86
```
Luap99: just a heads up https://github.com/containers/aardvark-dns/pull/695 will bump to rust 1.88

lsm5: ack, I'll rebase on that PR

lsm5:
> And that would allow us to have final status task, like "Total success". Without this in the github merge protection settings I need to set all exact task names

There's discussion and WIP happening in https://github.com/packit/packit-service/issues/1720 .

test/tmt/main.fmf
lsm5: The test will fetch rust version directly from Cargo.toml now.

lsm5:
I'd say this is ready for reviews. Will keep in draft until #695 is done as I rebased it already but this PR doesn't need to depend on it AFAICT.

@containers/netavark-maintainers @timcoding1988 PTAL

test/tmt/main.fmf
```diff
@@ -1,20 +1,23 @@
...
 
 /unit:
-    tag: [ no-rpm, unit ]
     summary: Unit tests
     test: make -C ../.. unit
     require:
         - cargo
+    adjust:
+        enabled: false
+        when: initiator is not defined or initiator != packit
```
Luap99, line 18: this is done so they do not run as part of Fedora CI/dist git setup I assume?

Unit tests are real test cases and they are faster than integration tests anyway so I think it would also make sense to run them as gating function. Not that I think this would catch anything particular but if we have this we should run this, better safe than sorry here.

Note I do agree we must skip validate, because a new rust will likely cause new validation errors due new lint rules.

.packit.yaml
```diff
@@ -113,15 +115,37 @@ jobs:
...
         - artifacts:
           - type: repository-file
             id: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/repo/centos-stream-$releasever/rhcontainerbot-podman-next-centos-stream-$releasever.repo
 
+  # Validate test (no RPM build needed)
+  - job: tests
+    trigger: pull_request
+    identifier: validate
+    skip_build: true
+    notifications: *test_failure_notification
```
Luap99: Do you actually want this? If this is the primary CI than it is the author/maintainers job to look at the tests results. There is no point in spam pinging you for this.

If I or another maintainer sees general packit problems we can ping you then directly so you do not drown in noise.

test/tmt/main.fmf
```diff
@@ -27,3 +27,18 @@
     adjust:
         duration: 10m
         when: arch == aarch64
+
+/msrv:
+    summary: Minimum Supported Rust Version (MSRV) check
+    test: |
+        MSRV=$(grep '^rust-version' ../../Cargo.toml | cut -d'"' -f2)
```
Luap99, line 34: good idea!

test/tmt/main.fmf
```diff
@@ -27,3 +27,18 @@
...
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain "$MSRV"
+        source $HOME/.cargo/env
```
Luap99, line 36: generally not a huge fan of this as hoc installs due to flake risk but I guess we will see how it turns out. If it starts flaking to often we can consider another approach.

Luap99 (commented):
LGTM overall, just some minor comments.

timcoding1988:
> Tests failed. @containers/packit-build please check.

small. am i missing something here, looks like tests all passed here.  also the total success check seems still visible from here. maybe remove it possible ?
other than that, all LGTM. thank you for setting this up !

.packit.yaml
lsm5: will remove all notifications in the upcoming update to this PR. Thanks!

test/tmt/main.fmf
lsm5: Are you ok leaving this for followup? IIUC, this would involve cloning the source tree in some form at downstream CI and we can't just use the `-tests` subpackage. That would need some more changes and downstream testing.

test/tmt/main.fmf
lsm5: ok, leaving this as is then.

lsm5:
> small. am i missing something here, looks like tests all passed here. 

I think Packit does a few retries so that's why you saw the failure notification without failures. But I'm removing the notifications in the upcoming force push, so this should be a non-issue.

> also the total success check seems still visible from here. maybe remove it possible ?

I think that needs to be removed from branch protection in repo settings which @Luap99 will take care of.

> other than that, all LGTM. thank you for setting this up !

Thanks

test/tmt/main.fmf
Luap99: ack sure absolutely not blocking, lets get it working as main upstream CI first and foremost right now

Luap99: Not sure if we would actually want to do this but AFAIK some packages run unit tests as part of the rpm spec `%check` which would have access to the full source tree? Anyhow not important right now.

Luap99:
> I think that needs to be removed from branch protection in repo settings which @Luap99 will take care of.

correct as long as a task is required in merge rule settings GH will display it

test/tmt/main.fmf
lsm5: > Not sure if we would actually want to do this but AFAIK some packages run unit tests as part of the rpm spec `%check` which would have access to the full source tree?

Ack yes, but that runs as part of rpmbuild / koji build which would be not so nice imo. I'm thinking either package the source as a `-src` subpackage or fetch TMT plan from the upstream tag. Let's see later ...

lsm5:
`three subnets, one container on two of the subnets, network connect` flaked once. Is that known? 

I've rebased only on main, so this is good to go without waiting for #695 .

Luap99:
> three subnets, one container on two of the subnets, network connect flaked once. Is that known?

I have no observed any failures with cirrus in the past, but overall a lot of these test things are timing related so some flakes can be possible I think.

Is there a way like in cirrus where I can see *all* runs from the repo so I can see the history which failed run you meant, i.e.
https://cirrus-ci.com/github/containers/aardvark-dns/main

Also the cirrus task always displayed at the top if they were a rerun and the previous log, I don't packit marks rerun in any way do they?

lsm5:
> Is there a way like in cirrus where I can see _all_ runs from the repo so I can see the history which failed run you meant, i.e. https://cirrus-ci.com/github/containers/aardvark-dns/main
> 
> Also the cirrus task always displayed at the top if they were a rerun and the previous log, I don't packit marks rerun in any way do they?

Hmm, don't know offhand. I'll ask around and file a request if they don't have it already.

Luap99:
https://dashboard.packit.dev/projects/github.com/containers/aardvark-dns?forge=github.com&namespace=containers&repo-name=aardvark-dns

seems to show a list of PRs and its copr builds links but not the testing farm links AFAICT

lsm5:
I think https://dashboard.packit.dev/jobs/testing-farm will show all jobs, but that's currently a nightmare to navigate, no filtering from what I can tell.  

cc @lecrisUT see https://github.com/containers/aardvark-dns/pull/697#issuecomment-4424142445

LecrisUT:
Unfortunately I do not have any better solution and web-dev is not in my skillset to navigate https://github.com/packit/dashboard/. If there are ideas of FOSS dashboards that we could host, we can explore that as well

Luap99:
> I think https://dashboard.packit.dev/jobs/testing-farm will show all jobs, but that's currently a nightmare to navigate, no filtering from what I can tell.

right. One of the nice thing of the cirrus UI is I can filter this by branch. So I can only look at results on main because it should not contain broken code so any failure there is most likely a flake and could be worth reporting/tracking/fixing...
And of course all of this is available via API as well so I have some basic scripts to check flake rates via that.

On PRs contributors push broken code all the time so a failing status on a PR does not mean anything really.

So overall I think it would be a nice feature request if we could have a webpage+API to get a list of all created copr and testing-farm jobs filtered by repo and branch if possible.

lsm5:
@containers/netavark-maintainers I think we're good to go here unless anything unaddressed. PTAL.

Luap99:
yes logic wise this seems good to me, I like to have https://github.com/containers/aardvark-dns/pull/695 merged first before I update the merge protection settings here

ashley-cui:
LGTM

lsm5:
Rebased on the latest main. PTAL @containers/netavark-maintainers

Luap99 (approved):
LGTM,

branch protection is set to require this now
```
testing-farm:fedora-44-x86_64:msrv
testing-farm:fedora-44-x86_64:unit
testing-farm:fedora-44-x86_64:validate
testing-farm:fedora-44-x86_64:integration-fedora
testing-farm:fedora-44-aarch64:integration-fedora
```

When fedora upgrades to 45 we need to update this but one update every 6 months is not the end of the world.

lachmanfrantisek:
Hi!

Sorry for being late to the party... 

Looks like you wanted to see older Packit results..

2 ways:


1. If we are speaking about statuses for previous commits, GitHub status is always attached to commit where you can see previous runs:
obrazek

obrazek


2. Packit dashboard
On Packit dashboard, you can check PRs and take a look at the jobs under such PR:
obrazek


Would these work for you or there is something still missing?

(Also, in case of new (force)pushes, the previous jobs are cancelled if not finished to safe resources.)

Luap99:
Thanks!

> Would these work for you or there is something still missing?

Well the github UI is just terrible in the sense that it is really hard to find prior commits especially when being force pushed many times.

I see the packit dashboard has the info so that is good but it is really hard to navigate. i.e. I missed the drop down where need to select "Tests". I assumed it only showed the copr builds at first.

I think the missing gap here is I have no idea what build or pipeline id was used before. In cirrus I get a historically sorted view. 

- PR 697
  - <date time> latest push
    -- all job links under here 
  - <date time> prior push
    -- all job links under here  
  - <date time> push before that, etc...
 
Also when someone rerun a task in cirrus the new task would have a link at the top to the prior run. With packit it seems like it is deleting the jobs from github and creating new ones on rerun so unless you happen to safe the testing farm URL beforehand you need to go to a dashboard (the regular user will have no idea exists) to find all failed runs on the PR.

Anyhow just minor complains I guess, it is good the current dashboard carries all that info.

One maybe easy suggestion for you, please color the task status so it will be much easier to spot, `TestingFarmResult.passed` == green, `TestingFarmResult.canceled` == grey, `TestingFarmResult.error` == red, same for the copr status I guess.

Or add a status filter on that view as well.

> (Also, in case of new (force)pushes, the previous jobs are cancelled if not finished to safe resources.)

sure that is totally fine and expected behavior IMO

PR #706, opened 2026-06-09
Release v2.0.0

Luap99:


baude (approved):
LGTM; one nit is whether you want to explicitly say av 2.0 must be used with nv 2.0 .... you do mention the bit about podman.

mheon:
I don't think it's strictly necessary, given that we always require NV and AV be pinned together?

mheon (approved):
LGTM

Issue #248, opened 2022-10-31
dns request failed: request timed out


MartinX3:
I'm using arch linux, so the packages should have the newest version.

I'm using firewallD and rootless podman with netavark and aardvark-dns.

I understand, that rootless podman with netavark won't manage my firewallD, but I would like to know which rules I need to activate to avoid the spam in my journal.
And if the rule need to be in my loopback or network interface. (Also if it is enough to allow communication with the host instead of having an open port in the internet.

My dns resolver is systemd-resolved
```bash
$ ls -lha /etc/resolv.conf
lrwxrwxrwx 1 root root 39 31. Okt 10:22 /etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
```

My journal spam:
`aardvark-dns[6156]: 21433 dns request failed: request timed out`

The rootless container itself can ping to `google.com`.
I didn't test if they can ping to a container dns name.

MartinX3:
Ah, after turning /etc/resolv.conf again into a symlink of systemd-resolved
`sudo ln -rsf /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf`
the messages disappear.
I assume it wants now to call the localhost ip of the systemd-resolved instead of the network ip of the outside resolver.

I don't know how I could let this plugin verbose debug logging into the journal, but I'm glad that it is fixed now.

MartinX3:
And it's back
`dns request failed: request timed out`

Szwendacz99:
I detected the same problem on my raspberry pi with Fedora and rootless podman containers.

jmaris:
Having the exact same issue. DNS lookup is extremely unstable, only two thirds of  lookups work

MartinX3:
Now using a nextcloud container which communicates with a PostgreSQL and a LDAP container.
I access the nextcloud container with a nginx reverse proxy container from the internet.

1/3 of the nextcloud web interface results into a 502 bad gateway, because of the unstable DNS.
And my logs are getting spammed with
```
aardvark-dns[3617]: 50310 dns request failed: request timed out
```

I hope the next podman release will include the new network stack which hopefully fixes this issue.

baude:
It is impossible to help with these issues as reporters did not provide versions for podman, netavark, and aarvark-dns.  please provide as much relevant information as possible.

MartinX3:
podman: 4.3.1
netavark: 1.4.0
aardvark-dns: 1.4.0

baude:
[...]

PR #715, opened 2026-07-07
logging: Try all upstream targets before logging failure #248

hartmark:
I had AI help me fix this but it seems that with this fix aardvark doesn't spam log about the failures as the second upstream resolver try worked.

### DIFF
--- a/src/dns/coredns.rs
+++ b/src/dns/coredns.rs
@@ -270,11 +270,10 @@ impl CoreDns {
         proto: Protocol,
     ) {
         let mut timeout = DEFAULT_TIMEOUT;
-        // Remember do not divide by 0.
         if !nameservers.is_empty() {
             timeout = Duration::from_secs(5) / nameservers.len() as u32
         }
-        // forward dns request to hosts's /etc/resolv.conf
+        let mut resolved = false;
         for addr in nameservers {
             let (client, handle) = match proto {
                 Protocol::Udp => {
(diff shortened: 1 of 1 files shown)


Luap99:
https://github.com/containers/common/blob/main/CONTRIBUTING.md#sign-your-prs
we cannot accept contributions with unclear copyright status

hartmark:
> https://github.com/containers/common/blob/main/CONTRIBUTING.md#sign-your-prs we cannot accept contributions with unclear copyright status

I have attributed the change to myself as the change is quite trivial. And any random developer would come up with this fix with or without AI

src/dns/coredns.rs
```diff
@@ -270,11 +270,10 @@ impl CoreDns {
         proto: Protocol,
     ) {
         let mut timeout = DEFAULT_TIMEOUT;
-        // Remember do not divide by 0.
```
Luap99, line 273: this is removing a unrelated comment

src/dns/coredns.rs
```diff
@@ -270,11 +270,10 @@ impl CoreDns {
...
         if !nameservers.is_empty() {
             timeout = Duration::from_secs(5) / nameservers.len() as u32
         }
-        // forward dns request to hosts's /etc/resolv.conf
```
Luap99, line 277: same here the comment is deleted for no reason?

src/dns/coredns.rs
```diff
@@ -309,15 +308,21 @@ impl CoreDns {
                 }
             };
 
-            if let Some(resp) = forward_dns_req(client, req.clone()).await {
+            if let Some(resp) = forward_dns_req(client, req.clone(), addr).await {
                 if reply(&mut sender, src_address, &resp).is_some() {
-                    // request resolved from following resolver so
-                    // break and don't try other resolvers
```
Luap99, line 315: and again here the comment is removed

src/dns/coredns.rs
```diff
@@ -309,15 +308,21 @@ impl CoreDns {
...
             }
             handle.abort();
         }
+
+        if !resolved {
+            error!(
+                "dns request {} failed: all nameservers exhausted",
+                req.id
+            );
+        }
```
Luap99, line 325: I don't really see the point in this. this will still spam if you have just one server all or all fail...

I think having the failed requests on debug is fine, no need to spam the journal with the same errors over and over again.

Luap99 (changes requested):

Issue #707, opened 2026-06-10
error[E0425]: cannot find value `SYS_kcmp` in crate `libc`


osokin:
Hi,

I've tried to update `dns/aardvark-dns` port on FreeBSD 14.4 to the recent version 2.0, and received the following error message for a build:

```
error[E0425]: cannot find value `SYS_kcmp` in crate `libc`
  --> aardvark-dns-2.0.0/cargo-crates/inotify-0.11.2/src/fd_guard.rs:96:23
   |
96 |                 libc::SYS_kcmp,
   |                       ^^^^^^^^ not found in `libc`

For more information about this error, try `rustc --explain E0425`.
error: could not compile `inotify` (lib) due to 1 previous error
```

Previous version, 1.17.1, is building just fine.
Could you guide me how to fix the issue.

Thank you.

Luap99:
I was not aware this is packaged for freebsd. The compile error seems to come from the inotfy create, does kcmp exists on freebsd?

In any case it is likely best you report this directly to the inotify maintainer. To fix this for the time being you should be able to downgrade the inotfy version in Cargo.toml here

osokin:
> I was not aware this is packaged for freebsd. The compile error seems to come from the inotfy create, does kcmp exists on freebsd?

`kcmp(2)` is available:

```
KCMP(2)                   FreeBSD System Calls Manual                  KCMP(2)

NAME
     kcmp – compare two kernel objects

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <unistd.h>

     int
     kcmp(pid_t pid1, pid_t pid2, int type, uintptr_t idx1, uintptr_t idx2);
```
 
> In any case it is likely best you report this directly to the inotify maintainer. To fix this for the time being you should be able to downgrade the inotfy version in Cargo.toml here

The list of dependencies for aardvark-dns 2.0.0 has been updated with the `inotify-0.11.2` package, previous version (1.17.1) has no such dependency.

osokin:
[Reported](https://github.com/hannobraun/inotify-rs/issues/262)

Luap99:
Fixed in https://github.com/containers/aardvark-dns/pull/714

We likely do a new release in a few weeks here

PR #716, opened 2026-07-09
Fix build on FreeBSD

osokin:
These changes fix build on FreeBSD, #707


Luap99:
Thanks, we update deps automatically so I just merged https://github.com/containers/aardvark-dns/pull/714 already

Also for contributions we require a DCO signoff, see https://github.com/containers/common/blob/main/CONTRIBUTING.md#sign-your-prs

But since this fix is merged already I am going to close this

PR #723, opened 2026-07-29
fix new lint warnings in 1.97

Luap99:
A few minor lint fixes for useless_borrows_in_formatting, question_mark and byte_char_slices.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -119,19 +119,15 @@ impl DNSBackend {
     pub fn get_network_scoped_resolvers(&self, requester: &IpAddr) -> Option<Vec<IpAddr>> {
         let mut results: Vec<IpAddr> = Vec::new();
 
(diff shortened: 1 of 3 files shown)


src/backend/mod.rs
```diff
@@ -119,18 +119,16 @@ impl DNSBackend {
...
-            Some(nets) => {
-                for net in nets {
-                    match self.network_dns_server.get(net) {
-                        Some(resolvers) => results.extend_from_slice(resolvers),
-                        None => {
-                            continue;
-                        }
-                    };
-                }
+        {
```
mheon: Any reason for the extra braces here? I don't think this needs to be separated out

Luap99: Ah this was done with the lint auto fix rules, I guess they kept the AST in tact in case the "drop" scope matters which it does not here. I Clean it up

mheon (approved):
LGTM

PR #731, opened 2026-08-10
add syscall for close_range for non-glibc targets

### LINKED ISSUE #730: Musl: error[E0425]: cannot find function `close_range` in crate `libc`
compiling on musl fails with error:

```Rust
error[E0425]: cannot find function `close_range` in crate `libc`
  --> src/main.rs:37:28
   |
37 |     let _ = unsafe { libc::close_range(3, libc::c_uint::MAX, 0) };
   |                            ^^^^^^^^^^^ not found in `libc`

For more information about this error, try `rustc --explain E0425`.
error: could not compile `aardvark-dns` (bin "aardvark-dns") due to 1 previous error
```

due to `close_rang` (https://github.com/containers/aardvark-dns/commit/2599cdd1113501d2c3618f685151a3f3b38a1ada) being available only on glibc (https://github.com/rust-lang/libc/pull/3373).

Downstream bug: https://bugs.gentoo.org/980641

peppergrayxyz:
compiling on musl fails with error:
```
error[E0425]: cannot find function `close_range` in crate `libc`
```

use syscall for targets which do not provide a `close-range`-wrapper

Closes: #730

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,9 +32,19 @@ enum SubCommand {
     Version(version::Version),
 }
 
+#[cfg(any(all(target_os = "linux", target_env = "gnu"), target_os = "freebsd",))]
+unsafe fn close_range(first: libc::c_uint, last: libc::c_uint, flags: libc::c_int) -> libc::c_int {
+    unsafe { libc::close_range(first, last, flags) }
+}
+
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
LGTM, thanks

I guess we should at a musl build job at some point to avoid regressions like that

PR #739, opened 2026-09-01
packit: update target for validate/unit/msrv

Luap99:
We want fedora-latest-stable to test on the latest stable not the latest branched release, i.e. f45 right now. While it likely does not matter in practice having a more stable package set seems better.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -120,8 +120,8 @@ jobs:
     trigger: pull_request
     identifier: validate
     skip_build: true
-    targets:
-      - fedora-latest-x86_64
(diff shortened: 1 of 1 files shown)


Luap99:
@lsm5 PTAL

lsm5 (approved):
LGTM, we could use yaml anchors to reduce some dupes.

Luap99:
@mheon PTAL, should be good to merge now. I switched to an yaml anchor per @lsm5's comment

mheon (approved):
LGTM

